PR #1, opened 2024-07-19
Update package and tsconfig

nickbar01234:
# Description

* Remove `@types/firebase` as `firebase` comes with a type declaration.

Pasted Graphic

* Use path alias in `tsconfig.ts`


johnny-t06 (approved):
LGTM! Thanks Nick

PR #2, opened 2024-07-19
Interface and converter

lnguyen2693:


src/app/types/db.ts
```diff
@@ -0,0 +1,40 @@
...
+  onDuty: string[];
+  tags: Tags;
+}
+
+export interface Question {
+  title: string;
+  description: string;
+  public: boolean;
+  // should state be enum or boolean or string?
+  state: string;
```
nickbar01234: I think it should be an enum here.

```ts
enum QuestionState {
  ...
}
```

src/app/utils/firestore.ts
```diff
@@ -0,0 +1,62 @@
...
+      public: question.public,
+      state: question.state,
+      timestamp: question.timestamp,
+      group: question.group,
+      tags: question.tags,
+    };
+  },
+  fromFirestore(snapshot: QueryDocumentSnapshot, options: any): Question {
+    const data = snapshot.data(options);
+    if (!data) {
```
nickbar01234: When would there be a case where `data` is null? Maybe we can do

```ts
return { ... } as Question
```

src/app/types/db.ts
```diff
@@ -0,0 +1,40 @@
...
+  color: string;
+  colorFill: boolean;
+}
+
+export interface Course {
+  name: string;
+  location: string;
+  professors: string[];
+  students: string[];
+  TAs: string[];
```
nickbar01234: nit: `tas` instead of `TAs`

src/app/utils/firestore.ts
```diff
@@ -0,0 +1,62 @@
...
+};
+
+export const questionConverter = {
+  toFirestore(question: Question): DocumentData {
+    return {
+      title: question.title,
+      description: question.description,
+      public: question.public,
+      state: question.state,
+      timestamp: question.timestamp,
```
nickbar01234: I think timestamp should be generated on server side. If we generate it on client, it's possible to modify the clock

nickbar01234: https://stackoverflow.com/questions/69519447/how-to-get-server-timestamp-from-firebase-v9

src/app/utils/firestore.ts
```diff
@@ -0,0 +1,62 @@
+import { Course, Question } from "../types/db";
+import { DocumentData, QueryDocumentSnapshot } from "firebase/firestore";
+
+export const courseConverter = {
+  toFirestore(course: Course): DocumentData {
```
nickbar01234: I think having the type to be `Course` is a bit restricting for end-users, i.e they must always pass in the full-object. So `PartialWithFieldValue<Course>` might make more sense here. Similar comment for `Question`.

src/app/utils/firestore.ts
```diff
@@ -0,0 +1,62 @@
+import { Course, Question } from "../types/db";
```
nickbar01234: Side note, I think the converter should be in a dedicated folder. Maybe `services/firestore`?

nickbar01234 (changes requested):

src/app/utils/firestore.ts
johnny-t06: +1 I like that idea. 
Looks like `data` might be null if 1) the document doesn't exist or 2) the user is not authenticated/have the right perms
https://stackoverflow.com/questions/53607050/documentsnapshot-is-always-null-for-a-valid-document-reference

johnny-t06 (changes requested):
Looks great, I am just echoing Nick's comments

src/app/types/db.ts
```diff
@@ -0,0 +1,44 @@
...
+  name: string;
+  location: string;
+  professors: string[];
+  students: string[];
+  tas: string[];
+  onDuty: string[];
+  tags: Tags;
+}
+enum QuestionState{
+  "pending" = 1,
```
nickbar01234, line 31: nit: Should be capitalized

PR #3, opened 2024-07-23
add and update functions for course and question

lnguyen2693:


src/app/types/db.ts
```diff
@@ -20,20 +20,22 @@ interface TagOption {
...
 export interface Course {
   name: string;
+  courseId: string;
   location: string;
-  professors: string[];
-  students: string[];
-  tas: string[];
-  onDuty: string[];
-  tags: Tags;
+  professors?: string[];
```
nickbar01234: I don't think these fields should be optional.

src/app/utils/course.ts
```diff
@@ -0,0 +1,54 @@
...
+    const courseDoc: DocumentReference = doc(
+      db,
+      `courses/${course.courseId}`
+    ).withConverter(courseConverter);
+
+    const newCourse = {
+      name: course.name,
+      courseId: course.courseId,
+      location: course.location,
+      professors: course.professors || [],
```
nickbar01234: Should be `??` for default values.

src/app/utils/question.ts
```diff
@@ -0,0 +1,63 @@
+import { db } from "../../../firebase";
```
nickbar01234: nit: `question.ts` and `course.ts` can live in `services` maybe under a folder called `client`?

nickbar01234 (changes requested):
Let's think about `addCourse` and `addQuestion`. Some fields should be populated when you're filling out the form. Having `PartialWithFieldValue` allows caller to choose what values they can pass in. Additionally, should we return the doc after adding it?

src/app/utils/course.ts
```diff
@@ -0,0 +1,54 @@
+import { db } from "../../../firebase";
```
nickbar01234: Maybe we can include a `getCourse` utility too?

src/app/utils/question.ts
```diff
@@ -0,0 +1,63 @@
...
+      public: question.public,
+      state: QuestionState.Pending,
+      timestamp: question.timestamp,
+      group: question.group,
+      tags: question.tags,
+    };
+
+    const questionDoc = await addDoc(questionsColection, newQuestion);
+
+    const questionId = questionDoc.id;
```
nickbar01234: Do we need a separate field to keep track of the ID when a doc has an ID?

src/app/types/db.ts
```diff
@@ -20,20 +20,22 @@ interface TagOption {
...
+  students?: string[];
+  tas?: string[];
+  onDuty?: string[];
+  tags?: Tags;
 }
-enum QuestionState{
-  "pending" = 1,
-  "resolved",
+export enum QuestionState{
+  Pending = 1,
```
nickbar01234: nit: All capital case

src/app/utils/question.ts
lnguyen2693: The ID is stored there so we know which question the user is interacting with in the UI

src/app/services/client/course.ts
```diff
@@ -0,0 +1,87 @@
+import { db } from "../../../../firebase";
+import { Course } from "../../interfaces/db";
```
nickbar01234: nit: Use path alias, i.e `@interfaces/db`

src/app/services/firestore.ts
```diff
@@ -9,6 +9,7 @@ export const courseConverter = {
   toFirestore(course: PartialWithFieldValue<Course>): DocumentData {
     return {
       name: course.name,
+      courseId: course.courseId,
```
nickbar01234: Similar to `Question`, I don't think we need to store `courseId` here.

nickbar01234 (approved):
Overall changes looks good, just one small comment on the ID

src/app/services/firestore.ts
```diff
@@ -28,7 +28,7 @@ export const courseConverter = {
       location: data.location,
       professors: data.professors,
       students: data.students,
-      tas: data.TAs,
+      tas: data.tas,
       onDuty: data.onDuty,
       tags: data.tags,
       // questions: data.questions,
```
johnny-t06, line 34: I think that we should keep the id (and in Course type) when retrieving from Firestore so our data is consistent on client and db.

src/app/services/client/course.ts
```diff
@@ -0,0 +1,89 @@
...
+  updateDoc,
+} from "firebase/firestore";
+
+type addCourse = Pick<Course, "name" | "location" | "professors"> &
+PartialWithFieldValue<Course>
+
+export const addCourse = async (
+  course: addCourse,
+  courseID: string,
+) => {
```
johnny-t06, line 19: Shouldn't we let Firestore handle unique courseid generation instead? I think we can take advantage of the addDoc here instead of the setDoc

src/app/services/client/course.ts
```diff
@@ -0,0 +1,89 @@
...
+      tas: course.tas ?? [],
+      onDuty: course.onDuty ?? [],
+      tags: course.tags ?? {},
+    };
+
+    await setDoc(courseDoc, newCourse);
+
+    console.log("New course created: ");
+    return courseDoc;
+  } catch (error) {
```
johnny-t06, line 40: If we get the reference on the addDoc `const newCourseDoc = await addDoc(courseDoc, newCourse);`

We can return `{...newCourse, courseId: newCourseDoc.id}`

src/app/services/firestore.ts
johnny-t06: Similar note, shouldn't our Question type have a uid as well?

nickbar01234: I believe `id` is built into Firebase documents.

```ts
const doc = await getDoc(<some Id>) ;
if (doc.exists()) {
   doc.id // accessible
}
```

src/app/services/client/course.ts
nickbar01234: If I understand you correctly, do you mean keep ID in the interface, but do not add it to documents on Firebase?

PR #4, opened 2024-07-24
Provide default data

nickbar01234:
# Description

Implement a context manager to provide `Course` and `Questions`.

## Additional Documentations

The current implementation only provides mock-data. We still need to hook up to the database and add listeners to handle data-change. I plan to add the logic in `useCourseLoader` and `useQuestionsLoader`.

### DIFF
--- a/src/app/hooks/utils/useLoadingValue.tsx
+++ b/src/app/hooks/utils/useLoadingValue.tsx
@@ -1,0 +1,30 @@
+import React from "react";
+
+export enum State {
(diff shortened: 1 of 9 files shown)


src/app/hooks/utils/useLoadingValue.tsx
johnny-t06, line 1: very coool!

johnny-t06 (approved):
LGTM, great work Nick!

PR #7, opened 2024-08-01
change converter, utility functions, and context

lnguyen2693:
Changes in utility functions:
- Takes in a full object
- Return an Identifiable type object (include an id, extend of each interface)

Changes in context:
- utilize type IdentifiableObject

### DIFF
--- a/src/app/services/client/course.ts
+++ b/src/app/services/client/course.ts
@@ -1,35 +1,26 @@
 import { db } from "../../../../firebase";
-import { Course } from "@interfaces/db";
 import { courseConverter } from "../firestore";
 import {
   deleteDoc,
   doc,
   DocumentReference,
   getDoc,
-  PartialWithFieldValue,
   setDoc,
   updateDoc,
 } from "firebase/firestore";
-import { defaultCourse } from "services/utils/defaultValue";
 import { IdentifiableCourse } from "@interfaces/type";
 
-
-type addCourse = Pick<Course, "name" | "location" | "professors"> &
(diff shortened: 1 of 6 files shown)


src/app/services/client/course.ts
```diff
@@ -5,85 +5,60 @@ import {
...
 } from "firebase/firestore";
+import { defaultCourse } from "services/utils/defaultValue";
+import { IdentifiableCourse } from "@interfaces/type";
+
 
 type addCourse = Pick<Course, "name" | "location" | "professors"> &
-PartialWithFieldValue<Course>
+  PartialWithFieldValue<Course>;
+
+export const addCourse = async (course: Course, courseID: string) => {
```
johnny-t06: Small nit, in terms of code consistency: should we also pass in `course: IdentifiableCourse` like our getCourse?

src/app/services/client/course.ts
```diff
@@ -5,85 +5,60 @@ import {
...
-export const addCourse = async (
-  course: addCourse,
-  courseID: string,
-) => {
-  try {
-    const courseDoc: DocumentReference = doc(
-      db,
-      `courses/${courseID}`
-    ).withConverter(courseConverter);
+  // const newCourse = defaultCourse(course);
```
johnny-t06: I think we can remove this

johnny-t06 (approved):
LGTM, good job guys! Just left a small comment, also we can remove the old code in the util functions if its not needed.

PR #9, opened 2024-08-03
Johnny-Won/frontend revamp and UserSessionContext

johnny-t06:
## Description
1. Created a UserSessionContext to provide a IdentifiableUser type to the app along with authentication details
2. Refactored our v1 components to support the new firebase backend and utilities. Added more fields to the User type + util functions + converter. 

## Notes
You can generally skim through the component usage because we are going to implement a new front end anyways. This was to show that the logic of our types and utility function works as intended, so we can reference back later on.
The v1 components live in @components/OldComponents and we can slowly sweep it out as we implement the new front end. 

I think you guys can mostly comment on the new types and the context!


src/app/utils/index.ts
```diff
@@ -0,0 +1,55 @@
+import { IdentifiableQuestion } from "@interfaces/type";
+import { FieldValue, Timestamp } from "firebase/firestore";
+
+export const trimName = (name: string) => {
+  const [firstName, lastName] = name.split(" ");
+  return firstName + " " + lastName + ".";
```
nickbar01234: Shouldn't we index only the first character for the last name?

Issue #10, opened 2024-08-05
Header, Tabs Navigation, and Error/Message Popup


johnny-t06:
Shared Components!
## Description
Implement the following shared components as it appears throughout many pages : header, bottom tabs selector, and Error/Message Alert!

![image](https://github.com/user-attachments/assets/0cb46aa8-1e27-48a6-b575-325f5c7de804)

PR #17, opened 2024-08-08
Johnny/navigation header

johnny-t06:
## Description
- Set up the file structure for the pages
- Implement bottom nav bar and header with mui components
- Added Tailwind Support 
- 
## Issue
Solve #10 except Error/Message popup

## Test
Mobile View visit to the pages `private/course`, `private/board`, `private/queue`, and `private/profile`
Edit: Pages are now nested into `private/course/[courseId]`
## Notes
The header is kinda janky but I spent a lot of time trying to figure out how to make the component accept:
- centered and left - aligned texts
- left/right icons and buttons

Getting the combinations of these were kinda tricky but maybe there's a better way? 

Also I didn't implement the error popup because it'll be a little easier to test as we implement our client rendering components + its an edge case popup


src/app/private/board/page.tsx
```diff
@@ -0,0 +1,22 @@
+import Header from "@components/Header";
+import { Button, Typography } from "@mui/material";
+
+const Page = () => {
+  return (
+    <div>
+      <Header
```
wkim10: I think it might be better to put Header components into layout.tsx

src/app/private/board/page.tsx
```diff
@@ -0,0 +1,22 @@
...
+        title="Board"
+        rightIcon={
+          <Button style={{ textTransform: "none" }}>
```
wkim10: nit: is there a way to use tailwind for this? i tried adding a class and doesn't seem to work cause mui but maybe there's a workaround?

src/app/globals.css
```diff
@@ -16,15 +16,15 @@
   }
 }
 
-body {
+/* body {
```
wkim10: nit: remove comment if not needed

src/app/components/NavBar.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import BottomNavigation from "@mui/material/BottomNavigation";
+import { BottomNavigationAction } from "@mui/material";
+import { SvgIconComponent } from "@mui/icons-material";
+
+interface INavBarItem {
+  label: string;
+  icon: React.ReactElement<SvgIconComponent>;
+  href: string;
+  // active: boolean;
```
wkim10: nit: do we need this + what is active for?

src/app/private/board/page.tsx
wkim10: oh wait unless we don't want it to display for nested pages/layouts, i just re-checked the figma

src/app/private/layout.tsx
```diff
@@ -0,0 +1,42 @@
+import HomeOutlinedIcon from "@mui/icons-material/HomeOutlined";
+import StickyNote2OutlinedIcon from "@mui/icons-material/StickyNote2Outlined";
+import PeopleAltOutlinedIcon from "@mui/icons-material/PeopleAltOutlined";
+import PersonOutlineOutlinedIcon from "@mui/icons-material/PersonOutlineOutlined";
+import { NavBarContainer } from "@components/container";
+import HeaderContainer from "@components/container/HeaderContainer";
```
wkim10: nit: remove import if we don't need

src/app/private/board/page.tsx
johnny-t06: Yeah I couldn't think of an elegant solution for a Layout component because each Tab (profile, board, queue, profile) and their nested pages have different headers

src/app/components/NavBar.tsx
johnny-t06: I was thinking to display a different (active) icon when the user is on that page/tab, but we can handle that next time.

src/app/private/board/page.tsx
wkim10: yeah makes sense i think it looks good for now 👍

wkim10 (approved):
good job as always mr pm

src/app/components/Header.tsx
```diff
@@ -0,0 +1,39 @@
...
+        {leftIcon}
+        <Typography
+          variant="h6"
+          component="div"
+          className={`flex-grow ${alignCenter ? "text-center" : "text-start"}`}
+        >
+          {title}
+        </Typography>
+
+        {alignCenter && !rightIcon ? <Box className="w-[40px]" /> : rightIcon}
```
wkim10, line 25: a little confused about this logic, if alignCenter=false and rightIcon isn't given will rightIcon be null/undefined? i tried it and seems to be working but a little confused

src/app/hooks/useAuth.tsx
```diff
@@ -15,7 +15,7 @@ const useAuth = () => {
   const onSignIn = useCallback(() => {
     const unsub = onAuthStateChanged(auth, (user) => {
       if (user) {
-        router.push("/course");
+        router.push("/private/course/cs160");
```
wkim10, line 18: this is just temporary right? like we would normally just push to /course?

src/app/components/Header.tsx
```diff
@@ -0,0 +1,39 @@
+import {
+  AppBar,
+  Box,
+  Button,
```
wkim10: nit: remove unneeded imports

src/app/components/Header.tsx
johnny-t06: My thought process is this: 
- If alignCenter == true and there is no right icon, we need to return an empty div/box.
- if alignCenter == true and there is a right icon, we can render the right icon.
- if alignCenter == false and there is no icon, it will go to the else statement of the ternary and try to render `{rightIcon}` which would be nothing.

src/app/hooks/useAuth.tsx
johnny-t06: Yes, thanks for pointing that out! We will change this once the userContext #9 is merged in.

PR #18, opened 2024-08-10
Add question board

nickbar01234:
# Description

Implement component to display question board.

image

I skipped some part of the Figma since we don't have all the necessary information yet, for example `UserProvider`. Additionally, there's no logic for the buttons.


src/app/interfaces/db.ts
```diff
@@ -41,7 +41,7 @@ export interface Question {
   state: QuestionState;
   timestamp: Date;
   group: string[];
-  tags: TagOption[];
+  tags: string[];
```
nickbar01234, line 48: Removed this since it's not necessary for the newest design.

src/app/course/[courseId]/board/page.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const addOrRemoveTopic = (topic: string) => {
+    setSelectedTopics((topics) => {
+      if (topics.includes(topic)) {
+        return topics.filter((other) => other !== topic);
+      } else {
+        return [...topics, topic];
+      }
+    });
+  };
+
```
Comment (johnny-t06) [inline src/app/course/[courseId]/board/page.tsx]: If the user selects the "All" topic, I think we should deselect the other topics.
And if the user selects some other topic, we can deselect the All topic.

src/app/course/[courseId]/board/page.tsx
```diff
@@ -0,0 +1,93 @@
...
+            sx={{
+              flexShrink: 0,
+              padding: "8px 16px",
+              color: "#1D192B",
+              whiteSpace: "nowrap",
+              minWidth: "content",
+              display: "flex",
+              alignItems: "center",
+              justifyContent: "center",
+              backgroundColor: selectedTopics.includes(topic) ? "#D7E3F8" : "",
```
Comment (johnny-t06) [inline src/app/course/[courseId]/board/page.tsx]: Weird bug with the topic background color selection. The topic states should be fine since the checkmark is being rendered conditionally, possibly an under the hood bug with the backgroundColor prop?

https://github.com/user-attachments/assets/233d9630-fdf2-4a5d-83c0-23df2f4a2935

johnny-t06 (changes requested):
The Board looks great Nick! Since the new changes to main, the path to the course is `private/course/[courseId]` so we can move your changes into there. Also we now have tailwind support for the mui components if you'd like to change that!

src/app/course/[courseId]/board/page.tsx
Comment (nickbar01234) [reply src/app/course/[courseId]/board/page.tsx]: It's lagging behind because there's a hover effect on MUI button. If you're on a non-mobile screen, it looks fine. I wonder if it works as expected on actual phone

Comment (johnny-t06) [reply src/app/course/[courseId]/board/page.tsx]: Yea, seems like this a :hover bug with mui button with some browsers. (omitted the mui github issue link so our work doesn't auto-link to their issue). There doesn't seem to be a fix with responsive view on chrome, but this should work on the actual phone!

src/app/context/UserSessionContext.tsx
```diff
@@ -0,0 +1,115 @@
...
+    error: null,
+  });
+  const router = useRouter();
+
+  React.useEffect(() => {
+    const unsub = onAuthStateChanged(auth, async (firebaseUser) => {
+      if (firebaseUser) {
+        const currUser = await getUser(firebaseUser?.uid);
+        setUser(currUser);
+        setSession({ isAuthenticated: true, isLoading: false, error: null });
```
johnny-t06, line 51: I wanted to `router.push("private/course/...")` but this ended in a infinite routing loop. We can explore that soon

johnny-t06:
LGTM, made some front-end tweaks to match figma and also messed around with the usersession. Ran into some auth routing infinite loop bug when we route on successful login.

PR #21, opened 2024-08-17
Display History Board

nickbar01234:
# Description

* Fix hover button select styling ([commit](https://github.com/johnny-t06/officeminutesv2/commit/dbc90f49b7b7191fabbfbd584625276f96897711))
* Fix bottom navigation overflow ([commit](https://github.com/johnny-t06/officeminutesv2/commit/62a3a29b04e4fa298b8d11eef6219d9a75542f26))
* Implement history board ([commit](https://github.com/johnny-t06/officeminutesv2/commit/fc5b6d1e73c4d17c0890df94d1d79e97c4fcab57))

image

## Additional Notes

* Current implementation doesn't distinguishes between public / private. Similar to #18, we don't have any logic yet.
* We also don't distinguish if a question belongs to the current user.

### DIFF
--- a/src/app/components/board/index.tsx
+++ b/src/app/components/board/index.tsx
@@ -1,13 +1,19 @@
 import { IdentifiableQuestions } from "@interfaces/type";
-import { Box } from "@mui/material";
+import { Box, Stack, ToggleButton, ToggleButtonGroup } from "@mui/material";
 import Question from "./Question";
+import { useOfficeHour } from "@hooks/oh/useOfficeHour";
+import React from "react";
+import CheckIcon from "@mui/icons-material/Check";
+import { compareQuestions } from "@utils/index";
 
 interface BoardProps {
   questions: IdentifiableQuestions;
 }
 
-const Board = (props: BoardProps) => {
-  const { questions } = props;
+const SELECT_ALL = "All";
+
+const _Board = (props: BoardProps) => {
+  const questions = props.questions.sort(compareQuestions).reverse();
 
   if (questions.length === 0) {
     return (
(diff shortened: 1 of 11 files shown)


src/app/components/board/index.tsx
```diff
@@ -33,4 +39,101 @@ const Board = (props: BoardProps) => {
...
+      } else if (newTopics.includes(SELECT_ALL)) {
+        return [SELECT_ALL];
+      } else {
+        return newTopics;
+      }
+    });
+  };
+  return (
+    <>
+      <ToggleButtonGroup value={selectedTopics} onChange={addOrRemoveTopic}>
```
wkim10: Currently for some of the phones on my end (e.g. iPhone 12 Pro) if you click on all the options for CS160 it forces the entire screen to be scrollable in both the x and y directions. Could we make it so that the options are scrollable horizontally instead?
Screenshot 2024-08-18 at 4 57 51 PM

src/app/utils/index.ts
```diff
@@ -1,4 +1,5 @@
-import { IdentifiableQuestion } from "@interfaces/type";
+import { QuestionState } from "@interfaces/db";
+import { IdentifiableQuestion, IdentifiableQuestions } from "@interfaces/type";
 import { FieldValue, Timestamp } from "firebase/firestore";
```
wkim10, line 3: nit: remove unneeded import

src/app/private/course/[courseId]/board/page.tsx
```diff
@@ -3,7 +3,9 @@ import Board from "@components/board";
 import Header from "@components/Header";
 import { useOfficeHour } from "@hooks/oh/useOfficeHour";
 import { Box, Button, Typography } from "@mui/material";
-import CheckIcon from "@mui/icons-material/Check";
+import { addQuestion } from "@services/client/question";
```
Comment (wkim10) [inline src/app/private/course/[courseId]/board/page.tsx:6]: nit: remove unneeded import

src/app/components/board/index.tsx
```diff
@@ -33,4 +39,101 @@ const Board = (props: BoardProps) => {
   }
 };
 
+const Board = (props: BoardProps) => {
```
wkim10, line 42: Component separation is clean 🔥

src/app/components/board/index.tsx
nickbar01234: Are you zooming in on the screen?

image

I do think though it would be nice to add a little right padding on the header. What do you think?

wkim10: Yeah right padding sounds good 👍

Initially I'm not zoomed in, but if I click both Conceptual and HW Help it forces the whole page to zoom in. Even if you zoom out the page seems to still be off-centered (there's more space on the right compared to the left):

Screen Shot 2024-08-25 at 2 52 45 PM
Screen Shot 2024-08-25 at 2 52 58 PM

wkim10: Looks great now 🙌

wkim10 (approved):
Great job nick san

Issue #12, opened 2024-08-05
Home View


johnny-t06:
## Description
Implement this page to display the general course information!
![image](https://github.com/user-attachments/assets/14421718-f9b8-4068-b6e1-050377187aad)

PR #22, opened 2024-08-17
Johnny/home view

johnny-t06:
## Description
Implemented the main course page to display general information.
Implemented the tas page to display onDuty and offDuty tas.
Expanded onto the mui theme context/constructor with fonts and colors. 

## Notes
Encountered a dilenma with the /course page and the /tas . As we want to optimize with nextjs with SSR pages, I ran into a few problems:
- Mui Themes can only be accessed client side
- const {course} = useOfficeHour() is client side as it is a hook
- Since course.tas is a list of the uids of the tas, we must still fetch for IdentifiableUsers of the tas. 

So I created client components to handle the rendering and let the page handle the fetching. It does feel kind of backwards to fetch for the course when we have the useOfficeHour hook.
image
image

## Issue
Solve #12 , except the tas donot have the time logged/clocked in yet.


src/app/components/DisplayCourse.tsx
```diff
@@ -0,0 +1,134 @@
...
+              router.push(`${pathname}/tas`);
+            }}
+          >
+            <Typography variant="subtitle2">View All</Typography>
+          </Button>
+        </Box>
+        <Box sx={{ padding: "16px, 16px", height: "48%", overflow: "auto" }}>
+          <DisplayTas tas={tas} />
+        </Box>
+        <Button
```
wkim10, line 104: If there are a lot of TAs then this button hovers above the navbar like this (for iPhone 12 Pro):
Screenshot 2024-08-18 at 4 44 27 PM

And for iPhone SE it looks like this on my end:
Screenshot 2024-08-18 at 4 47 25 PM

src/app/private/course/[courseId]/page.tsx
```diff
@@ -1,33 +1,34 @@
 import Header from "@components/Header";
```
Comment (wkim10) [inline src/app/private/course/[courseId]/page.tsx:1]: nit: For this page and the page.tsx for TAs, in cases where the content on the page doesn't fill up the whole page (e.g. if there are only a few TAs so the list is short) it still seems to be able to be scrolled, maybe we can only allow it to scroll when the content fills up the page

src/app/components/DisplayCourse.tsx
```diff
@@ -0,0 +1,134 @@
+"use client";
```
wkim10, line 1: The components are 🔥

src/app/components/DisplayCourse.tsx
wkim10: finished!

src/app/components/DisplayCourse.tsx
wkim10: thank you won!

src/app/private/course/[courseId]/page.tsx
Comment (wkim10) [reply src/app/private/course/[courseId]/page.tsx:1]: great job johnny!

johnny-t06:
Note: TA's page has a default height for OnDuty and OffDuty TAs

locmai1:
LGTM! great job guys!

PR #23, opened 2024-08-27
Form to join queue and edit submission

lnguyen2693:
Screenshot 2024-08-27 at 12 24 41 PM

In this PR:
- Form to join queue that can be reused to edit submission
- Button to open form and create new question

Note:
- The current implementation does not keep the order of tags in the form
- The chosen tags buttons in the form still need styling


package.json
```diff
@@ -12,13 +12,14 @@
...
-    "@mui/material-nextjs": "^5.16.6",
     "@mui/icons-material": "^5.16.7",
     "@mui/material": "^5.16.7",
+    "@mui/material-nextjs": "^5.16.6",
     "firebase": "^10.12.3",
     "next": "14.2.4",
     "react": "^18",
-    "react-dom": "^18"
+    "react-dom": "^18",
+    "react-icons": "^5.3.0"
```
nickbar01234: Do we need react-icons?

src/app/components/queue/CreateQuestion.tsx
```diff
@@ -0,0 +1,47 @@
+import { Box, Button } from "@mui/material";
```
nickbar01234, line 1: Oops, duplicated with my implementation [here](https://github.com/johnny-t06/officeminutesv2/pull/25/files#diff-b99b071e626569a6c49511173bb0aec380c0f8dbf6fcf4a14f4d8820110de56bR63-R82)

nickbar01234: ~May be we can remove `private/course/[courseId]/queue.tsx` and `CreateQuestion.tsx` from the PR, since it's not directly related to the task?~

src/app/components/queue/form/Tags.tsx
```diff
@@ -0,0 +1,132 @@
...
+    let tag = tags.options.filter((o) => change[o.choice]);
+    updateQuestionTags(tagsKey, tag);
+  };
+  return (
+    <Box>
+      <FormControl required={tags.required}>
+        <FormLabel id={tagsKey}>{tagsKey}</FormLabel>
+        <Box>
+          {tags.options.map((o) => (
+            <Checkbox
```
nickbar01234, line 55: Don't forget `key={o}`

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+                control={
+                  <Checkbox
+                    onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+                      setQuestionPublic(event.target.checked)
+                    }
+                    checked={questionPublic}
+                  />
+                }
+                label={
+                  <Box>
```
nickbar01234: I think this can use some more spacing.

image

src/app/components/queue/CreateQuestion.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { FaArrowRight } from "react-icons/fa";
```
nickbar01234: You can get the same arrow from MUI icon library. I would say we should use that since that's what the Figma is based off.

src/app/components/queue/CreateQuestion.tsx
```diff
@@ -0,0 +1,47 @@
...
+import QuestionForm from "./form/QuestionForm";
+import { defaultQuestion } from "@utils/index";
+
+const CreateQuestion = () => {
+  const triggerButton = (
```
nickbar01234, line 7: Why don't we just render the button directly in `CreateQuestionForm`?

src/app/components/queue/CreateQuestion.tsx
nickbar01234: Edit: I see what you're doing with `CreateQuestion.tsx`. We probably should use this instead of my implementation.

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+  currentQuestion: IdentifiableQuestion;
+}
+
+const QuestionForm = (props: QuestionFormProps) => {
+  const { triggerButton, title, currentQuestion } = props;
+  const [openForm, setOpenForm] = React.useState(false);
+  const ohContext = React.useContext(officeHourContext);
+  const userSessionContext = React.useContext(UserSessionContext);
+
+  const [question, setQuestion] = React.useState(currentQuestion.title);
```
nickbar01234: I don't think we need to keep track of the state for each field. We can do simply

```tsx
const [question, setQuestion] = React.useState(currentQuestion)
```

nickbar01234: I don't think we need to keep track of the state for each field. We can do simply

```tsx
const [question, setQuestion] = React.useState(currentQuestion)
```

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+      return [...arr, ...curr];
+    }, []);
+
+    const tagsValidate = Object.keys(questionTags).filter(
+      (k) => ohContext.course.tags[k].required && questionTags[k].length === 0
+    );
+
+    if (question !== "" && description !== "" && tagsValidate.length === 0) {
+      if (title === "Join queue") {
+        createQuestion(
```
nickbar01234: We will probably use `react-hook-form` or a similar library for field validation, so we can omit this check

nickbar01234: We will probably use `react-hook-form` or a similar library for field validation, so we can omit this check

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  const [question, setQuestion] = React.useState(currentQuestion.title);
+  const [description, setDescription] = React.useState(
+    currentQuestion.description
+  );
+  const [questionPublic, setQuestionPublic] = React.useState(
+    currentQuestion.questionPublic
+  );
+
+  const [questionTags, setQuestionTags] = React.useState<
```
nickbar01234: Spoke offline, we should derive `questionTags` from `currentQuestion`.

src/app/components/queue/CreateQuestion.tsx
```diff
@@ -0,0 +1,47 @@
...
+        zIndex: 99,
+      }}
+    >
+      <Box display="flex" justifyContent="center" alignItems="center">
+        <FaArrowRight />
+      </Box>
+      <Box display="flex" justifyContent="center" alignItems="center">
+        Join queue
+      </Box>
+    </Button>
```
johnny-t06, line 38: In Nick's PR, he used the Fab which is just slightly cleaner. Which one do you think we should use?[Here]( https://github.com/johnny-t06/officeminutesv2/pull/25/files#r1778895020) @lnguyen2693

src/app/components/queue/form/QuestionForm.tsx
johnny-t06: I personally haven't used react-hook-form, would love to hear more thoughts. @locmai1 @wkim10

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+              label="Description"
+              placeholder="Add description"
+              focused
+              multiline
+              rows={4}
+              value={description}
+              onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+                setDescription(event.target.value);
+              }}
+            ></TextField>
```
johnny-t06: nit: the textfields can be self closing

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+interface QuestionFormProps {
+  // button to open the form
+  triggerButton: JSX.Element;
+  title: string;
+  // current state of question
+  currentQuestion: IdentifiableQuestion;
+}
+
+const QuestionForm = (props: QuestionFormProps) => {
```
johnny-t06, line 29: Are we looking to use the same component for editting the form? We will have to conditionally render

src/app/components/queue/form/QuestionForm.tsx
johnny-t06: I agree!

src/app/components/queue/form/QuestionForm.tsx
johnny-t06: I am following the logic, and it should populate the questionTags for either the "create new form" and "edit form". It is a little tricky though, I'm not sure if we can improve it for our types.

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+  const updateQuestionTags = (tagsKey: string, newTags: TagOption[]) => {
+    setQuestionTags({ ...questionTags, [tagsKey]: newTags });
+  };
+
+  const trigger = React.cloneElement(triggerButton, {
+    onClick: () => {
+      setOpenForm(true);
+    },
+  });
+
```
johnny-t06, line 63: It might be cleaner to let the client of this component to handle the onClick, we may want to add more features for the onClick that could differentiate for "creating a question" and "editing a form".

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+    } else {
+      console.log("Required fields not filled")
+      // TODO(lnguye2693) - Display error
+    }
+
+    setOpenForm(false);
+  };
+  return (
+    <Box>
+      {trigger}
```
johnny-t06, line 103: image
Can we add more padding to the form to line up with the header like the figma? The form feels a little close to the edges

johnny-t06: Also can add some marginLefts for the Good, Confused, I'm Lost selections.
Like Nick said, I think we should make the "Post to board" smaller

src/app/components/queue/form/Tags.tsx
```diff
@@ -0,0 +1,132 @@
...
+        Object.hasOwn(allQuestionTags, tagsKey) &&
+        allQuestionTags[tagsKey].includes(o)
+      ) {
+        init[o.choice] = true;
+      } else {
+        init[o.choice] = false;
+      }
+    });
+    return init;
+  });
```
johnny-t06, line 37: This logic took me a little bit to understand, but it works!

src/app/context/UserSessionContext.tsx
```diff
@@ -26,7 +26,7 @@ interface IUserSessionContext {
   onSignOut: () => Promise<void>;
 }
 
-const UserSessionContext = React.createContext<IUserSessionContext>(
+export const UserSessionContext = React.createContext<IUserSessionContext>(
```
johnny-t06, line 30: We can omit this export, because we actually have the hook useUserSession at the bottom of this file. You can replace it in Questionform.tsx

src/app/private/course/[courseId]/queue/page.tsx
```diff
@@ -1,4 +1,7 @@
+'use client'
```
Comment (johnny-t06) [inline src/app/private/course/[courseId]/queue/page.tsx:1]: do we need the useclient change for the component usage?

johnny-t06 (commented):
Looks great so far, a few styling changes requested and then we can re-review.

src/app/components/queue/form/QuestionForm.tsx
lnguyen2693: okk I fixed it now the client can add more features to their button while it can still be used to open the form when passed into the QuestionForm

src/app/context/UserSessionContext.tsx
lnguyen2693: oh.. I'm not quite sure how to to use it tbh

src/app/components/queue/form/QuestionForm.tsx
lnguyen2693: yeah we do have to conditionally render it in QuestionForm, not sure if there's any better way tho

src/app/components/queue/CreateQuestion.tsx
lnguyen2693: Looked into it! I don't think it makes any difference here tho

lnguyen2693:
Screenshot 2024-12-24 at 1 59 14 PM


good to merge? @johnny-t06

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,244 @@
...
+  Box,
+  Button,
+  Checkbox,
+  Drawer,
+  FormControl,
+  FormControlLabel,
+  FormLabel,
+  TextField,
+} from "@mui/material";
+import { IoMdClose } from "react-icons/io";
```
wkim10: nit: it seems like we're only using react-icons here, is there any chance we can find the corresponding icon in MUI for consistency?

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,244 @@
...
+          timestamp,
+          author ? [author] : [],
+          tagsArr,
+          ohContext.course.id
+        );
+      }
+      // if title === edit submission
+    } else {
+      console.log("Required fields not filled");
+      // TODO(lnguye2693) - Display error
```
wkim10, line 94: Is the plan to implement this on a future ticket?

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,244 @@
...
+                  color: "text.primary",
+                  textTransform: "none",
+                  borderRadius: 50,
+                  width: 170,
+                  marginTop: 2,
+                  marginBottom: 2,
+                }}
+              >
+                <FileUploadOutlinedIcon></FileUploadOutlinedIcon>
+                <Box sx={{ margin: 0.5, marginLeft: 1.5 }}>Upload image</Box>
```
wkim10, line 194: Will the "Upload image" button be implemented on a future ticket?

src/app/components/queue/form/QuestionForm.tsx
lnguyen2693: Ideally now, I'm looking around to see what we are using to display error messages, but I also thought of using Toast

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const trigger = React.cloneElement(triggerButton, {
+    onClick: () => {
+      setOpenForm(true);
+      if (typeof triggerButton.props.onClick === 'function') {
+        triggerButton.props.onClick();
+      }
+    },
+  });
+
+  const joinQueue = () => {
```
wkim10: I'm not sure what's going on but when I click "Join queue" I'm getting an error saying "Cannot read properties of null (reading 'toDate')". I did some console logging and it seems that the question I just submitted has a timestamp of "null" instead of the current time, which I'm confused about since you set the timestamp in this function. Did you ever run into this issue before?

src/app/components/queue/form/QuestionForm.tsx
wkim10: Sorry for the late reply, I've also never used react-hook-form but if you think it'll be cleaner than manually checking for empty fields then I'm down to try that out

src/app/components/queue/form/QuestionForm.tsx
wkim10: Yeah I think makes sense to conditionally render! Since they're almost identical

src/app/components/queue/form/QuestionForm.tsx
wkim10: But if I reload the page after I get the error, my question shows up and the queue looks great

wkim10 (commented):
Really great work Linh the UI looks basically identical to the Figma 🙌. Just wanted to double check, but it looks like there's a little more stuff to complete which you left in your TODO comments, is the plan for the ticket to include those changes too or just the Join queue functionality for now?

src/app/components/queue/form/QuestionForm.tsx
lnguyen2693: woahh I didn't know about the bug before you found out here. Thanks. It's an onSnapshot bug, fixed it!

src/app/components/board/Question.tsx
```diff
@@ -114,7 +114,7 @@ const Question = (props: QuestionProps) => {
               color="#43474E"
             >
               <Typography sx={{ fontWeight: 500, fontSize: 14 }}>
-                {tag}
+                {tag.toString()}
```
johnny-t06: image
Since tag is type TagOption as an object, a toString() would not work here for rendering or as a key.

src/app/components/board/index.tsx
```diff
@@ -58,7 +58,7 @@ const Board = (props: BoardProps) => {
       return questions;
     } else {
       return questions.filter((question) =>
-        question.tags.find((tag) => topics.includes(tag))
+        question.tags.find((tag) => topics.includes(tag.toString()))
```
johnny-t06: Here as well for the TagOption toString()

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,244 @@
...
+  };
+
+  const trigger = React.cloneElement(triggerButton, {
+    onClick: () => {
+      setOpenForm(true);
+      if (typeof triggerButton.props.onClick === "function") {
+        triggerButton.props.onClick();
+      }
+    },
+  });
```
johnny-t06, line 62: Interesting!

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+  const joinQueue = () => {
+    const timestamp = Timestamp.now();
+    const author = userSessionContext.user?.tufts_username;
```
johnny-t06: Let's favor name for now so TAs can easily call out students.

src/app/components/queue/form/QuestionForm.tsx
johnny-t06: Because of time, we can keep this and and disallow form submission based on this check for now. 
Unless you think it's necessary @nickbar01234 ?

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,244 @@
...
+              }}
+            />
+
+            {/* TODO(lnguyen2693) - Add index for tags, sort and display them 
+            in an order */}
+            {Object.keys(ohContext.course.tags)
+              .sort(
+                (a, b) =>
+                  ohContext.course.tags[a].priority -
+                  ohContext.course.tags[b].priority
```
johnny-t06: I do think priority is a good idea, but I'm having a hard time seeing how TAs would determine priority of tags on creation. 

Let's sort by alphabetical of the TagOptions.choice for now.

src/app/components/queue/form/QuestionForm.tsx
johnny-t06: Lets omit this for a later update because we won't need it for the mvp.

src/app/private/course/[courseId]/queue/page.tsx
```diff
@@ -21,6 +22,7 @@ const Page = () => {
         title={course.name}
         alignCenter
       />
+      <CreateQuestion></CreateQuestion>
```
Comment (johnny-t06) [inline src/app/private/course/[courseId]/queue/page.tsx]: nit: You can self close `<CreateQuestion/>`

src/app/utils/index.ts
```diff
@@ -68,6 +69,40 @@ export const formatTimeDifference = (
...
+    description: description,
+    questionPublic: questionPublic,
+    state: QuestionState.PENDING,
+    timestamp: timestamp,
+    group: group,
+    tags: tags,
+  };
+
+  addQuestion(question, courseId);
+};
```
johnny-t06: Lets have this in the `/client/question.ts`. Or you can make a new folder api folder?

src/app/services/client/question.ts
```diff
@@ -18,7 +18,7 @@ import { IdentifiableQuestion, IdentifiableQuestions } from "@interfaces/type";
 
 type AddQuestion = Pick<
   Question,
-  "title" | "description" | "public" | "group" | "tags"
+  "title" | "description" | "questionPublic" | "timestamp" | "group" | "tags"
```
johnny-t06: Here we can omit timestamp to favour a serverTimeStamp() in addQuestion() in `client/question.ts` to rely on Google's generation rather than the client.

src/app/utils/index.ts
johnny-t06: Looks like some types are missing from imports.

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -0,0 +1,244 @@
...
+    );
+
+    if (question !== "" && description !== "" && tagsValidate.length === 0) {
+      if (title === "Join queue") {
+        createQuestion(
+          question,
+          description,
+          questionPublic,
+          timestamp,
+          author ? [author] : [],
```
johnny-t06: In your createQuestion Api, you have a group field rather than author. Also Questions should store user uids in the group field rather than their names.

johnny-t06 (changes requested):
Thanks for the hard work! I left comments on things to omit and change, let me know if you have questions!

src/app/components/board/Question.tsx
```diff
@@ -50,7 +50,7 @@ const Question = (props: QuestionProps) => {
         alignItems="center"
       >
         <Avatar sx={{ bgcolor: theme.palette.primary.main }}>
-          {trimName(users[0].name[0]) ?? ""}
+          {trimName(users && users[0]?.name[0]) ?? ""}
```
nickbar01234: Why are we indexing `name[0]`? I think this can be simplified to just `trimName(users[0]?.name ?? "")`?

nickbar01234: Actually the implementation for `trimName` already accounts for undefined, so can simply do `trimName(users[0])` and update the type definition

src/app/components/board/Question.tsx
```diff
@@ -105,7 +105,7 @@ const Question = (props: QuestionProps) => {
         <Box display="flex" columnGap="16px" rowGap="8px" flexWrap="wrap">
           {question.tags.map((tag) => (
             <Box
-              key={tag}
+              key={tag.toString()}
```
nickbar01234: Is this valid in React?

nickbar01234: Actually no, since object to string returns `[object Object]`. We can do `JSON.stringify(Object.values(tag).sort())`.

src/app/components/queue/CreateQuestion.tsx
lnguyen2693: Cloning a button that the client passes in allows the client to turn any button they have into a trigger to open the form. If we render the button directly in CreateQuestionForm, we can only have 1 design for the openForm button. I want to use the form for both creating and editing questions, and the 2 buttons might have different designs. That's why I allow the client to give their own button and I will clone it in the form to add the openForm functionality.

src/app/components/queue/form/QuestionForm.tsx
lnguyen2693: The `group` field is a list of strings that has its first string being the `author`, so the `[author]` should be fine. If there is anyone wanting to join the group, we will just add their credential to the list.

I think the idea before was we used a Tufts username to identify the author, and it should be unique for each user. That's why I used author (Tufts username) here! Also because of that, in your comment [here](https://github.com/johnny-t06/officeminutesv2/pull/23/files#r1902484689), did you mean we just use the uid there? Since I don't think what stored in the `author` field will show up on the UI

src/app/components/queue/form/QuestionForm.tsx
lnguyen2693: For now, I add it to match the design. What I was considering was the priority is automatically given to the tags based on the order in which they are created.  And the TAs can change it if they want. What do you think? I think it would make more sense if we keep it this way.

johnny-t06: Editing and deleting the tags would add complexity to maintain the priority of each tag in the backend. I think it might be too much logic for the mvp right now.

Do you think revisit later on and sort by alphabetical right now?

johnny-t06: Editing and deleting the tags would add complexity to maintain the priority of each tag in the backend. I think it might be too much logic for the mvp right now.

Do you think revisit later on and sort by alphabetical right now?

src/app/components/queue/form/QuestionForm.tsx
johnny-t06: Oops sorry, yeah the group field takes in a list of strings so passing a list makes sense! 

But I do think we should pass in the user's uid because

1. While Tufts username should be unique, it's up to tufts to generate that and for our auth to extract the username from the email (which isn't being accomplished yet). I don't think we use tufts_username in the app yet

2. The board and question component right now expects that group has a list of user uids

lnguyen2693: 🫡🫡

src/app/components/queue/form/QuestionForm.tsx
lnguyen2693: Have this block so the cloned button still has features of the original button that the client give us. Explained why I didn't render the trigger button in the form [here](https://github.com/johnny-t06/officeminutesv2/pull/23#discussion_r1903142881)

lnguyen2693:
:// sorry about the tags on the board I fixed it.
Screenshot 2025-01-05 at 7 05 36 PM

sort tags alphabetically
Screenshot 2025-01-05 at 7 06 08 PM

src/app/hooks/oh/useQuestionsLoader.tsx
```diff
@@ -42,12 +42,14 @@ export const useQuestionsLoader = (props: UseQuestionsLoaderProps) => {
         questionConverter
       ),
       (snapshot) => {
-        let questionsList: IdentifiableQuestions = [];
-        snapshot.docs.map((doc) => {
-          questionsList.push({ id: doc.id, ...doc.data() });
-        });
+        if (!snapshot.metadata.hasPendingWrites) {
```
johnny-t06: I'm reading up the usage of this check. Its a battle of consistency vs speed. This is core to our concurrency handling so I am pinging everyone @lnguyen2693 @nickbar01234 @wkim10 @locmai1 

I am testing the join group functionality by joining a public question in `board/[uid]`.

If we do perform `!hasPendingWrite` check:
- **The main issue**: Front end state (local to the component, not the OHContext) is updated but if you navigate with the header and back to `board/[uid]` -- the component state reverts/does not persist.

- **Why?**: This is because the updateDoc causes two writes (a [local firebase cache](https://firebase.google.com/docs/firestore/query-data/listen) and server write). Both local and server writes should be the same data (under no race conditions). Local write will immediately invoke snapshot listeners ("latency compensation"). `QuestionLoader/onSnapshot` does not update OHContext because the server write is not fully committed yet due to the `!hasPendingCheck`. When the server write is fully committed, you would expect the onSnapshot to fire again. However since local cache data and server data is the same, Firebase does not send another snapshot therefore OHcontext is stale and then the UI is stale. (The [link]( https://firebase.google.com/docs/firestore/query-data/listen) is helpful in understanding)

What if we don't perform the check?
 - The UI is instantly updated as I join the group. (and of course broadcasted to all other listeners). 
 - This is an optimistic writing approach. We assume that the data will always be written to the server without fail.
 - **Issue 1**: Firebase follows last-to-win write([1](https://stackoverflow.com/questions/47746100/how-reliable-is-firestore-as-an-offline-persistence-mechanism/47781502#47781502) and [2](https://www.youtube.com/watch?v=oDvdAFP6OhQ&t=344s)) which means if multiple users go offline, firebase just chooses the most recent write from all of the users when they come back online. The issue is that the users would see their offline change, come back online, then sync with the server data to see something maybe totally different. _I think this okay for now._
 - **Issue 2**: Concurrently, this would cause multiple UI changes possibly with some data overridden and intermediate states being rendered because pending writes are not being waited. 

We can discuss this more, but right now I think favouring faster UI data (without the check)  makes more sense because the two latter issues should not occur that often.

src/app/components/board/Question.tsx
johnny-t06: Fixed, this one's on me!

src/app/hooks/oh/useQuestionsLoader.tsx
nickbar01234: I believe the original reason why @lnguyen2693 added this flag was to account for the fact we can't Firebase `serverTimestamp` immediately on local write. Based on the [link](https://firebase.google.com/docs/firestore/query-data/listen) you provided, we should be observing the new document after the timestamp has been persisted.

> Then, each time the contents change, another call updates the document snapshot.

> Issue 1: Firebase follows last-to-win write([1](https://stackoverflow.com/questions/47746100/how-reliable-is-firestore-as-an-offline-persistence-mechanism/47781502#47781502) and [2](https://www.youtube.com/watch?v=oDvdAFP6OhQ&t=344s)) which means if multiple users go offline, firebase just chooses the most recent write from all of the users when they come back online. The issue is that the users would see their offline change, come back online, then sync with the server data to see something maybe totally different. I think this okay for now.

This is a fair concern! I believe this would be a problem depending on our write pattern. For example, we should be using `ArrayUnion` ([link](https://firebase.google.com/docs/firestore/manage-data/add-data#update_elements_in_an_array)), instead of doing `list: [...curList, newItem]`

> Issue 2: Concurrently, this would cause multiple UI changes possibly with some data overridden and intermediate states being rendered because pending writes are not being waited.

Do you have a specific instance in mind or a sequence of events that would lead to that?

johnny-t06: Ah great catch on the server timestamp! Are you preferring to perform the `!hasPendingWrites` check?
Otherwise, do you think a fallback like `timestamp: data.timestamp ?? new Timestamp()` is sufficient?

> This is a fair concern! I believe this would be a problem depending on our write pattern. For example, we should be using ArrayUnion ([link](https://firebase.google.com/docs/firestore/manage-data/add-data#update_elements_in_an_array)), instead of doing list: [...curList, newItem]

I see, will change!

> Do you have a specific instance in mind or a sequence of events that would lead to that?

The ArrayUnion solves the scenario that would affect us the most (with joining group). Previously, if two users joined at the same time -- one user could override the other user being added to the group because we handled joining group with list replacement rather than union-ing/mutating. 

Another scenario I was thinking: 
1. User A makes class red. (Local write, server write ongoing)
2. User B makes class blue, at the same time. (Local write, server write ongoing).
3. User B server write finishes first, makes user A and B's class is now blue.
4. User A server write finishes, makes user A and B's classes red. 

Big picture wise though -- this wouldn't affect daily users that much. Students can't modify the same question. So it would mostly affect TAs, say editing tags or announcements. So maybe it isn't a big deal?

johnny-t06: >Otherwise, do you think a fallback like timestamp: data.timestamp ?? new Timestamp() is sufficient?

I believe the fallback would then be refreshed when the server write hits and actually generates the serverTimestamp().

src/app/components/queue/CreateQuestion.tsx
johnny-t06: Very cool!

johnny-t06:
I added a few changes to the branch:
1. Some trimUserName and sorting util functions
2. QuestionForm.tsx: combined the useStates to 1 Question state and 1 tags state.
3. Join and Leave group APIs with arrayUnion/Remove
4. Removed !hasPendingWrites and use client side Timestamp as fallback (subject to change)

src/app/hooks/oh/useQuestionsLoader.tsx
nickbar01234: > Otherwise, do you think a fallback like timestamp: data.timestamp ?? new Timestamp() is sufficient?

I think this depends on how much we trust the user. It's possible that someone changes the timezone or modify their OS clock to skip ahead of the queue.

> Another scenario I was thinking:
> User A makes class red. (Local write, server write ongoing)
> User B makes class blue, at the same time. (Local write, server write ongoing).
> User B server write finishes first, makes user A and B's class is now blue.
> User A server write finishes, makes user A and B's classes red.

If users override the same field, this should be the expected behavior right? Maybe a different edge case is

1. User A and B starts with `{ color: "RED", courseName: "Algorithms" }`.
2. User A changes to `{ color: "BLUE", courseName: "Algorithms" }`.
3. User B changes to `{ color: "RED", courseName: "Data Structure" }`.
4. User A submit changes
5. User B submit changes without seeing (4).
6. We end up with `{ color: "RED", courseName: "Data Structure" }`, in which case this is definitely a bug we need to solve.

johnny-t06: > I think this depends on how much we trust the user. It's possible that someone changes the timezone or modify their OS clock to skip ahead of the queue.

I tested this by setting the fallback `Timestamp.fromDate(new Date("2025-01-06T00:00:01")),` and having questions later than 00:00:01. You would expect it to shift the questions order but it doesn't. With console.logs, the server write (with the serverTimestamp) is immediately served to the client after the 2025-01-06T00:00:01 log. So there's no weird re-ordering. Also, the fallback is for the client themselves and the serverTimestamp is sent to the backend so it shouldn't be possible to skip the queue. 

> We end up with { color: "RED", courseName: "Data Structure" }, in which case this is definitely a bug we need to solve.

Yes that makes sense. An update to multiple fields is considered as a single write. We could 1) use Firebase transactions 2) Add lastupdatedfields in course and question documents -- to force a refresh when local data somehow is not in sync or 3) combine both?

Edit: Yeah concurrent editing of the same field shouldn't be a problem

nickbar01234: > Yes that makes sense. An update to multiple fields is considered as a single write. We could 1) use Firebase transactions 2) Add lastupdatedfields in course and question documents -- to force a refresh when local data somehow is not in sync or 3) combine both?

Perhaps we should not override the entire object every time and only send to firebase the fields that have been modified. If 2 users simultaneously override the same field, then I don't think it's a weird behavior to one have win over the other nondeterministically. The alternative I can think of is having a special role that only permits specific TA to to edit fields.

Issue #27, opened 2024-11-30
[Optimization] Remove unnecessary reads/unsubscribe 


locmai1:
Things to note for reads:

- Unsubscribe for users
- Unsubscribe correctly for Questions
- Unsubscribe correctly for Courses

Issue #26, opened 2024-11-27
[Bug] [P1] Firebase snapshot getting excessive reads


locmai1:
2 problems at hand (from Nick):

- The auth recursively runs
- Even within navigating between /[courseId]/<some-nested-page> the hooks fire again
- onAuthStateChange calls setUser is recursively run -> the hook triggers when user changes

PR #28, opened 2024-12-01
Fix client routing

nickbar01234:
# Description

Currently, every navigation action between tabs result in a full-rerender from the root component. Specifically, the current implementation used in MUI does not respect how next preserve context between client navigation. The implication means that:
* The listeners are triggered multiple times.
* A full rerender defeats the purpose of using server / client components.

The fix follows from this [discussion](https://github.com/mui/material-ui/issues/26655). Also took this opportunity to remove `pnpm-lock`, which I accidentally added in a previous PR :)


johnny-t06 (approved):
LGTM, great find! I just removed the user dependency from the useEffect in userSessionContext during my testing, to prevent the recursive firing you mentioned. The contexts are preserved!

johnny-t06:
Also fixes #27 and #26 by removing redundant context subscribes

PR #29, opened 2024-12-01
Course/Question Loader Context Unsub

johnny-t06:
# Description

When we were exploring why our firebase reads were reaching high reads, we looked into the two Firebase snapshot locations. Now that #28 fixes the unnecessary context renders, we will still bring some of the changes for cleaner code.

React's useEffect cleanup expects a function rather than an invoke. We can cleanup and reset the unsubscribe ref.

### DIFF
--- a/src/app/context/UserSessionContext.tsx
+++ b/src/app/context/UserSessionContext.tsx
@@ -8,7 +8,7 @@ import {
   signInWithPopup,
   signOut,
 } from "firebase/auth";
-import React from "react";
+import React, { useRef } from "react";
 import { auth } from "@project/firebase";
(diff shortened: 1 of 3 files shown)


nickbar01234:
Is a function different than an invoke?

johnny-t06:
My understanding is that React expects a function to cleanup whenever the dependencies of the useEffect changes. If provided an invoke in the return, it calls the function immediately and unsubs from the listeners rather than waiting for a dependency change.

nickbar01234:
Just came across this [post](https://stackoverflow.com/questions/71149561/react-unsubscribe-from-rxjs-subscription-in-useeffect), so this makes sense to me

PR #31, opened 2024-12-15
Course navigation sidebar

locmai1:
# Description

Added functionality to the MenuIcon. Only existing right now on 4 pages so far:

- `private/course`
- `private/course/[courseId]`
- `private/course/[courseId]/queue`
- `private/course/[courseId]/profile`

## Issues

None :)

## Screenshots

Behavior on `officeminutesapp@gmail.com` account (has all courses cs160 & cs40):

image

Behavior on `locmaiworks@gmail.com` account (has only cs40):

image

Should appear as Drawer upon clicking MenuIcon on the Header

## Test

Test all courses in sidebar + manage courses on each of the 4 routes:

- Manage Courses will route to `/private/course/[courseId]/profile/edit` (not implemented yet but I think it's supposed to be like that according to Figma?)
- Course will route to according course and if on `/private/course/cs160` and press `cs40`, it should route to `/private/course/cs40`

## Possible Downsides

None :) (that I can think of)

## Additional Documentations

Not sure that I did this the best way: used memo and a provider to reduce props drilling and should only re-render on change to userSession doc. Please let me know!


src/app/components/Sidebar.tsx
```diff
@@ -0,0 +1,157 @@
...
+      : `${pathname}/${courses[0].id}`;
+
+    router.push(`${basePath}/profile/edit`);
+    closeSidebar();
+  };
+
+  const formattedCourseName = (course: IdentifiableCourse) => {
+    const formattedCourseId = course.id
+      .toUpperCase()
+      .replace(/([a-zA-Z]+)(\d+)/, "$1 $2");
```
nickbar01234, line 59: Double checking - should there be a space?

src/app/components/Sidebar.tsx
```diff
@@ -0,0 +1,157 @@
...
+  };
+
+  const handleManageCoursesClick = () => {
+    const isCourseRoute = /\/[a-zA-Z]+\d+/.test(pathname);
+
+    const basePath = isCourseRoute
+      ? pathname.match(/(.*\/[a-zA-Z]+\d+)/)?.[0]
+      : `${pathname}/${courses[0].id}`;
+
+    router.push(`${basePath}/profile/edit`);
```
nickbar01234: Can you clarify why we need to dynamically figure out the base path?

src/app/components/Sidebar.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const { courses: courseIds = [] } = user || {};
+  const router = useRouter();
+  const pathname = usePathname();
+
+  React.useEffect(() => {
+    const loadCourses = async () => {
+      const allCourses = await getCourses();
+      const filteredCourses = allCourses
+        .filter((course) => courseIds.includes(course.id))
+        .map((course) => course);
```
nickbar01234: I don't think we need the map here?

src/app/components/Sidebar.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const { isOpen } = useSidebarStateContext();
+  const { closeSidebar } = useSidebarActionsContext();
+  const { user } = useUserSession();
+  const { courses: courseIds = [] } = user || {};
+  const router = useRouter();
+  const pathname = usePathname();
+
+  React.useEffect(() => {
+    const loadCourses = async () => {
+      const allCourses = await getCourses();
```
nickbar01234: We probably should do the filtering on database side.

src/app/components/Sidebar.tsx
```diff
@@ -0,0 +1,157 @@
...
+import theme from "theme";
+import { getCourses } from "@services/client/course";
+import { useUserSession } from "@context/UserSessionContext";
+import { useRouter } from "next/navigation";
+import { usePathname } from "next/navigation";
+import { IdentifiableCourse } from "@interfaces/type";
+import MenuButton from "./buttons/MenuButton";
+import ModeEditOutlinedIcon from "@mui/icons-material/ModeEditOutlined";
+
+const Sidebar = React.memo(() => {
```
nickbar01234, line 24: Why do we need both `Sidebar` and `SidebarContext`? Can the state not live inside this component?

nickbar01234:
Double checking, we need to place this in `/board` page too right?

src/app/components/Sidebar.tsx
nickbar01234: Ah I see it's because of the `Header` component.

src/app/components/Sidebar.tsx
locmai1: Yeah I think so, according to Figma at least:

image

nickbar01234: I see some wireframe has it together - might be worth confirming with Fa and Sarah
image

src/app/components/Sidebar.tsx
locmai1: /profile lives inside /[courseId] and sidebar is also accessible in /course so if in /course, it needs to go to /[courseId]/profile/edit so i'm thinking we take the first of courseIds from userSession as part of the basePath. lmk what you think

src/app/components/Sidebar.tsx
locmai1: yes, we don't. good catch!

src/app/components/Sidebar.tsx
locmai1: is that making a server component?

nickbar01234: The query to firebase should already filter to the course ids we're interested in.

src/app/components/Sidebar.tsx
nickbar01234: Can't we always do 

```ts
route.push(`/private/${currentCourseId}/profile/edit
```
As a separate note, it feels weird that we have profile nested in `/[courseId]`.

src/app/components/Sidebar.tsx
locmai1: I thought about this too and I reasoned that:

- MUI Drawer component needs to in higher-order component as it covers the screen
- We would reduce prop drilling in the component tree 
- We would also reduce renders of Sidebar on different pages since it's static to userSession

locmai1:
> Double checking, we need to place this in `/board` page too right?

According to the Figma, I don't think so? there's no Menu Icon

src/app/services/client/course.ts
```diff
@@ -57,6 +57,15 @@ export const getCourses = async () => {
   return coursesDocs;
 };
 
+export const getCoursesByIds = async (
+  courseIds: String[]
+): Promise<IdentifiableCourse[]> => {
+  const courses = courseIds.map((id) => getCourse(id));
```
nickbar01234: I think firestore should support filtering https://cloud.google.com/firestore/docs/query-data/queries#in_not-in_and_array-contains-any

src/app/context/SidebarContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+      </SidebarActionsContext.Provider>
+    </SidebarStateContext.Provider>
+  );
+};
+
+export const useSidebarStateContext = () =>
+  React.useContext(SidebarStateContext);
+
+export const useSidebarActionsContext = () =>
+  React.useContext(SidebarActionsContext);
```
johnny-t06, line 51: What's the reasoning behind two contexts? The boolean and actions are so tied together. 

But we should probably merge in #33 first and then update this context.

src/app/private/course/[courseId]/page.tsx
```diff
@@ -19,11 +19,7 @@ const Page = async (props: PageProps) => {
   return (
     <div>
       <Header
-        leftIcon={
-          <IconButton edge="start">
-            <MenuIcon />
-          </IconButton>
-        }
+        leftIcon={<MenuButton isEdge={true} />}
```
Comment (johnny-t06) [inline src/app/private/course/[courseId]/page.tsx]: small nit: Can just use `isEdge`

johnny-t06 (commented):
LGTM, looks great! We should wait for #33 to merge so the contexts can be updated as well.

PR #33, opened 2024-12-16
React context to zustand migration

locmai1:
# Description

- Context to zustand migration
- Removal of `context` directory + `hooks/oh` as everything is done with useStores
- Just using a zustand store (no provider) as it is a singular global state

## Issues

- Removing the bloat of context providers

## Screenshots

N/A

## Test

Ran a normal workflow of local server and checked all functionality

## Possible Downsides

- can't use next/router on state functions as zustand is outside the component tree
- not sure if entire application is covered in this whoops

## Additional Documentations

[Zustand Nextjs documentation](https://zustand.docs.pmnd.rs/guides/nextjs)


src/app/stores/useOfficeHourStore.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export const useOfficeHourStore = create<OfficeHourState>((set) => ({
+  course: null,
+  questions: [],
+  isLoading: true,
+  error: null,
+  setCourse: (course) => set({ course }),
+  setQuestions: (questions) => set({ questions }),
+  setError: (error) => set({ error }),
+  setLoading: (loading) => set({ isLoading: loading }),
```
johnny-t06: I can't seem to think of a case for these setters? Since course and question data changes rely on the listeners, maybe we can omit them?

src/app/layout.tsx
```diff
@@ -19,9 +18,7 @@ export default function RootLayout({
     <html lang="en">
       <body className={inter.className} id="root">
         <div id="__next">
-          <UserSessionContextProvider>
-            <ThemeProviderWrapper>{children}</ThemeProviderWrapper>
-          </UserSessionContextProvider>
+          <ThemeProviderWrapper>{children}</ThemeProviderWrapper>
```
johnny-t06: Looking at https://zustand.docs.pmnd.rs/guides/nextjs#app-router
We need to provide the stores through contexts because of

> No global stores - Because the store should not be shared across requests, it should not be defined as a global variable. Instead, the store should be created per request.

src/app/components/board/index.tsx
```diff
@@ -41,8 +41,8 @@ const _Board = (props: BoardProps) => {
 
 const Board = (props: BoardProps) => {
   const { questions } = props;
-  const { course } = useOfficeHour();
-  const topics = [SELECT_ALL, ...Object.keys(course.tags).sort()];
+  const { course } = useOfficeHourStore();
+  const topics = [SELECT_ALL, ...Object.keys(course?.tags || {}).sort()];
```
johnny-t06, line 45: This is for some unrelated consideration, we should probably default tags for courses (in case some TA deletes all the tags).

src/app/layout.tsx
johnny-t06: Also, Zustand's error handling here reminded me of Next.js's error handling. It might be cleaner to just throw the error in our (Store initialization and Context) and fallback onto an error.tsx.  

Right now, OfficeHourState has an Error / Message state and I assume we were planning to conditionally render? Am I missing another usecase for the Error/Message state. 
@nickbar01234 what do you think?


> export const useCounterStore = <T,>(
>   selector: (store: CounterStore) => T,
> ): T => {
>   const counterStoreContext = useContext(CounterStoreContext)
> 
>   if (!counterStoreContext) {
>     throw new Error(`useCounterStore must be used within CounterStoreProvider`)
>   }
> 
>   return useStore(counterStoreContext, selector)
> }

src/app/stores/useUserSessionStore.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export const useUserSessionStore = create<UserSessionState>((set, get) => ({
+  user: null,
+  session: {
+    isAuthenticated: false,
+    isLoading: true,
+    error: null,
+  },
+  setUser: (user) => set({ user }),
+  setSession: (session) => set({ session }),
```
johnny-t06: Similarly to the OH store, I'm not sure if we need to expose any setters.

src/app/stores/useUserSessionStore.ts
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+      useUserSessionStore.getState().setUser(null);
+      useUserSessionStore.getState().setSession({
+        isAuthenticated: false,
+        isLoading: false,
+        error: null,
+      });
+    }
+  });
+}
```
johnny-t06: I'm not sure if this auth state listener would operate on the same context as the above above store. I think it should but maybe we can also be consistent with the OH store and have this auth listener as a initlistener state? What do you think?

johnny-t06 (commented):
Looking great, Zustand's is sick. I left a few comments about the structure of the stores.

 I tested the stores by updating the time fields of questions and it lived rendered. Course and Question data look correct. Sign in / Session still looks good.

src/app/stores/useUserSessionStore.ts
locmai1: It would definitely be nicer and cleaner but then we'd have to invoke it at some point before rendering. If you look at `/private/course/[courseId]/layout.tsx`:

```Javascript
React.useEffect(() => {
    const cleanup = initializeListeners(courseId);
    return () => cleanup();
  }, [courseId, initializeListeners]);
```

I reasoned that we should only call the listener when page is navigated to but with auth, I thought it'd be best if listener is invoked when the store is created.

Also, I had to make `layout.tsx` a client component but I don't know if there's any drawbacks in doing so.

 Please me know though, I could be missing something.

johnny-t06: Ah I see, making `/private/course/[courseId]/layout.tsx` a client component would increase page load times (because of more client sided hydration) to our core pages. 

I'm now leaning towards moving the OH initializing listeners to occur on store creation.

locmai1: Mmmm, then would we need to initializing all courses listeners on creation?

src/app/providers/OfficeHourProvider.tsx
```diff
@@ -0,0 +1,43 @@
+"use client";
```
nickbar01234, line 1: Why not consolidate both providers into one and have separate slices?

src/app/stores/officeHourStore.ts
```diff
@@ -0,0 +1,51 @@
+import { createStore } from "zustand";
+import { IdentifiableCourse, IdentifiableQuestions } from "@interfaces/type";
+import { doc, onSnapshot, collection } from "firebase/firestore";
+import { db } from "@project/firebase";
+import { courseConverter, questionConverter } from "@services/firestore";
+
+interface OfficeHourState {
+  course: IdentifiableCourse | null;
```
nickbar01234, line 8: What's the downside if we were to initialize this to an empty `IdentifiableCourse`? I'm not a fan of working with nulls; alternatively, it would be better if we ensure that the store can be used if and only if it has been initialized.

tailwind.config.ts
```diff
@@ -1,12 +1,16 @@
 import type { Config } from "tailwindcss";
 
 const config: Config = {
+  corePlugins: {
```
nickbar01234, line 4: What is this for?

locmai1: twas from combining the two different exports in tailwind.config.ts in font PR

PR #34, opened 2024-12-20
Profile Page

### LINKED ISSUE #19: [Feature] Profile, Add Classes, Remove Classes
## Description
This looks like a big ticket but do not be frightened! It is front-end heavy so it might take a bit.
Implement the main Profile page and the Edit Classes flow. This includes adding and removing classes.
We can omit the Get Help and Send Feedback to a later ticket!
## Notes
The profile route is currently `private/course/[courseId]/profile`, which is not the most intuitive. The motivation is that the `[courseId]/layout.tsx` provides the dynamic navigation bar such that each button will bring to the dynamic home, board, and queue. 
Clicking from profile to the other buttons will navigate back to the "correct" dynamic page, since the profile is dynamic as well. 

image

wkim10:
# Description

Finished profile page, including sending feedback to email functionality using SendGrid API (let me know if you need the API key)

## Issues

Resolve #19 

## Screenshots

Screenshot 2024-12-20 at 12 02 35 PM
Screenshot 2024-12-20 at 12 02 48 PM
Screenshot 2024-12-20 at 12 03 03 PM
Screenshot 2024-12-20 at 12 03 13 PM

## Test

Click on each of the subpages, try viewing on different phone dimensions, try sending different feedback/recommendations and see how it looks in the sent email.

## Possible Downsides

Still need to come up with answers to three of the questions in the "Get Help" subpage: "I don't think the queue is closed..."; "What happens if I miss..."; "What should I do if office..."

Also, haven't implemented the "Classes" subpage because not part of MVP :)

## Additional Documentations

Nothing else :)


src/app/api/send-feedback/route.client.ts
```diff
@@ -0,0 +1,14 @@
+export const sendFeedback = async (
+  recommendation: string,
+  feedback: string
```
nickbar01234: Can we have the body as object instead?

src/app/api/send-feedback/route.ts
```diff
@@ -0,0 +1,50 @@
+import sgMail from "@sendgrid/mail";
+import { NextRequest, NextResponse } from "next/server";
+
+sgMail.setApiKey(process.env.SENDGRID_API_KEY || "");
```
nickbar01234: If there's no API key, this will fail right?

src/app/private/course/[courseId]/profile/classes/page.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const Page = () => {
+  const router = useRouter();
+
+  return (
+    <div>
+      <Header
+        leftIcon={
+          <IconButton
+            onClick={() => {
```
Comment (nickbar01234) [inline src/app/private/course/[courseId]/profile/classes/page.tsx]: nit: Can be simplified to `onClick={router.back}`

src/app/api/send-feedback/route.client.ts
```diff
@@ -0,0 +1,11 @@
+export const sendFeedback = async (body: Record<string, any>) => {
```
nickbar01234: I would use an interface here.

nickbar01234:
Can you explain more on what the vision behind sendgrid is? Since this is internal, couldn't we also just store it in the database and aggregate it through an admin view?

wkim10:
> Can you explain more on what the vision behind sendgrid is? Since this is internal, couldn't we also just store it in the database and aggregate it through an admin view?

Yeah we definitely could. I was thinking along the lines that when sending feedback it'd be easy to just quickly check on email since it would send a notification, but I guess in the long term it'd definitely make more sense either to store it in the db and look at it through some kind of admin view or storing it into a spreadsheet/csv. I can take a look at + implement one of those options after we discuss.

src/app/services/client/feedback.ts
```diff
@@ -0,0 +1,64 @@
+import { IdentifiableFeedback, IdentifiableFeedbacks } from "@interfaces/type";
+import { db } from "../../../../firebase";
```
nickbar01234: We can use path alias here

src/app/private/course/[courseId]/profile/feedback/page.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const formControlLabelStyles = {
+    display: "flex",
+    flexDirection: "row-reverse",
+    alignItems: "center",
+    justifyContent: "space-between",
+    marginLeft: 0,
+    marginRight: 0,
+  };
+
+  const [recommendation, setRecommendation] = React.useState<number | null>(
```
Comment (nickbar01234) [inline src/app/private/course/[courseId]/profile/feedback/page.tsx]: I would opt in for an enum for better semantics? We can also construct an object which maps the recommendation to label, so the actual `FormControlLabel` is cleaner.

```ts
enum Recommendation {
   VERY_UNLIKELY;
   ...
}

type RecommendationLabel = Record<Recommendation, String>

const recommendations: RecommendationLabel = { Recommendation.VERY_UNLIKELY: "Very unlikely" };
```

I guess technically, we can also use a constant string enum

```ts
enum Recommendation {
  VERY_UNLIKELY="Very unlikely"
}
```

but the tradeoff is space, so I'm good with either.

Comment (nickbar01234) [reply src/app/private/course/[courseId]/profile/feedback/page.tsx]: Better yet,

```ts
type RecommendationLabel = { [key in Recommendation]: string };
```

enforces that all the enum value must be initializes

src/app/private/course/[courseId]/profile/feedback/page.tsx
```diff
@@ -0,0 +1,167 @@
...
+                backgroundColor: theme.palette.action.disabledBackground,
+                color: theme.palette.text.disabled,
+              },
+            }}
+            onClick={handleSendFeedback}
+            disabled={recommendation === null}
+          >
+            Send feedback
+          </Button>
+          {error !== "" ? (
```
Comment (nickbar01234) [inline src/app/private/course/[courseId]/profile/feedback/page.tsx]: I'm forgetting, is `error !== "" && <Component />` a valid syntax?

nickbar01234 (approved):
Approve with nits

src/app/private/course/[courseId]/profile/feedback/page.tsx
Comment (locmai1) [reply src/app/private/course/[courseId]/profile/feedback/page.tsx]: yeah it's valid. I think `!error && <Component />` also evaluates as well

PR #35, opened 2025-01-08
Show private questions in queue

lnguyen2693:
# Description

#### This PR fixes the following issue:
The queue doesn't display or count private questions, which can be confusing for the student if they see that it's their turn in the queue but the TAs go help other students

#### Example:
I add 5 questions to the queue:
1. Private 
2. Public
3. Public
4. Private
5. Public
--> There should be 5 questions in the queue. However, our current queue only shows 3 questions:
Screenshot 2025-01-08 at 3 24 46 PM

## Test

Add 5 questions to the queue. The queue displays the private questions and the board doesn't:
Screenshot 2025-01-08 at 3 28 31 PM
Screenshot 2025-01-08 at 3 29 00 PM


johnny-t06 (approved):
LGTM! Also, thinking we should do server sided filtering for students to be safe.

PR #36, opened 2025-01-08
TA View (Home, Announcements, Board, Queue)

johnny-t06:
# Description
Big PR for you guys!
In this PR, we separated pages and components into TA and Student Views:
**Home / Announcements**:
- TAs can add, edit, and delete announcements. + placeholder for no ann.
- The design choice for announcements was to embed into the course object. I didn't want to have another listener for an announcement sub collections. This does bring in finnicky api functions. Since the ann. are embedded, they do not have a uid unless we generate on frontend. Even then crud functions will still rely on Firebase's arrayUnion and arrayRemove (uid woudn't give an advantage here, so its omitted). So for edit, we remove and then add back the newAnn.

**Board**
- TAs do not see the Join Group / Leave Group button for the Questions

**Queue**
- Students now see students being helped + missing students
- TAs can now start helping, mark people as missing, and complete a question
- TAs can now open and close the queue

### Student View
Student View

### All populated
All populated

### Ta View
Ta View

### Start Helping
Start Helping

### Mark as Missing
Mark as Missing
Mark as Missing

### All populated
All populated

## Test
Will need to modify firebase a lot
1. Go to /cs160 as a student. (remove your id from cs160/tas if needed)
2. Make public and private questions
3. Check out board and queue to make sure ordering is good
4. Switch over to TA view by adding your id into cs160/tas
5. Click on questions on Queue page to mark as missing / start helping
6. Navigate back to queue to see if changes are reflected
7. You can also change Question states manually with firebase
8. Now check out announcements as a TA, edit, add, delete them. Mess with combinations of the buttons.
9. Can switch back to student to view announcements
 

## Possible Downsides

Announcement type and api structure is weird. Edit takes a remove and then a union. Favored this over setting up another listener that might cause more reads. Evens out with more writes rather than more reads. 
Also adding announcements, I couldn't use serverTimestamp() for createdAt because Firebase prohibits for arrays. I trust TAs for their client timestamp.


src/app/components/AnnouncementTile.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { Announcement } from "@interfaces/db";
+import { CustomModal } from "./CustomModal";
+interface AnnouncementTileProps {
+  announcement: Announcement;
+  isEdit: boolean;
+  startEdit?: boolean;
+}
+export const AnnouncementTile = (props: AnnouncementTileProps) => {
+  const { announcement, isEdit, startEdit = false } = props;
+  const [editing, setEditing] = React.useState<boolean>(false);
```
johnny-t06, line 26: isEdit toggles the editing mode for all tiles (to show Delete and Edit buttons)
startEdit is passed in when the TA creates an announcement 
editing is the boolean that allows the TA to type and update the announcement

src/app/components/DisplayAnnouncements.tsx
```diff
@@ -0,0 +1,109 @@
...
+      {
+        message: "",
+        createdAt: Timestamp.now(),
+      },
+    ]);
+    setIsEdit(true);
+  };
+  const onEditDone = () => {
+    if (isEdit && currAnnouncements.length > announcements.length) {
+      setCurrAnnouncements(announcements);
```
johnny-t06, line 40: If the new client sided announcement is not saved / posted, we can reset to the original state.

src/app/components/DisplayAnnouncements.tsx
```diff
@@ -0,0 +1,109 @@
...
+          isEdit={false}
+        />
+      ) : (
+        <Box sx={{ display: "flex", flexDirection: "column", gap: "12px" }}>
+          {sortAnnouncements(currAnnouncements).map((announcement, index) => (
+            <AnnouncementTile
+              key={index}
+              announcement={announcement}
+              isEdit={isEdit}
+              startEdit={announcement.message === ""}
```
johnny-t06, line 88: All real announcements will not have the message as  "" so only the newly created announcement triggers the typing mode of a tile.

src/app/components/StudentDisplayCourse.tsx
```diff
@@ -128,10 +105,10 @@ const DisplayCourse = (props: DisplayCourseProps) => {
           >
             Show Support to TAs
           </Typography>
-        </Button>
+        </CustomButton> */}
```
johnny-t06, line 108: Commenting this out to come back when we decide to implement the support wall

src/app/components/TADisplayCourse.tsx
```diff
@@ -0,0 +1,162 @@
...
+            display: "flex",
+            marginTop: "16px",
+          }}
+        >
+          <Typography variant="body2" color={theme.palette.text.primary}>
+            JCC 4th floor huddle room
+          </Typography>
+          <Typography variant="body2" color={theme.palette.text.primary}>
+            6 - 9 PM
+          </Typography>
```
johnny-t06, line 56: Oh, we should actually display these accordingly.

src/app/components/board/QuestionDetails.tsx
johnny-t06, line 1: This component honestly has a lot (probably too many) conditionals. 
For student view:
- Clicking a question in board
For TA view:
- Clicking a question in board
- Clicking a question in queue 
- Displayed when helping a question

All with dynamic buttons (color / text)

src/app/hooks/useCourseData.tsx
johnny-t06, line 1: Hook to mainly check if user is a TA but can also provide tas and students.

src/app/components/board/QuestionDetails.tsx
```diff
@@ -67,119 +86,205 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
...
-              {question.tags.map((tag) => (
-                <Box
-                  key={tag.choice}
-                  border={1}
-                  borderColor="#73777F"
-                  borderRadius="10px"
-                  paddingY="4px"
-                  paddingX="14px"
-                  color="#43474E"
+              <CustomButton
```
lnguyen2693, line 224: should we check if the user is TA here? I can still start helping others (and myself) even when I'm not a TA
Screenshot 2025-01-08 at 6 30 47 PM
Screenshot 2025-01-08 at 6 31 06 PM

lnguyen2693: Oops my bad I forgot to pull newest changes😭😭

lnguyen2693 (approved):
lgtm! You guys are GOAT
nit: I feel like the create/edit announcement is not very intuitive in terms of the placement of the buttons. For example, when I create a new ann. but then decide not to post it, I wasn't sure where to delete it besides clicking "post" and then "delete". Maybe we need a "cancel" button here?

src/app/components/board/QuestionDetails.tsx
wkim10: Yeahh there might be too many props, open to suggestions on a better way of approaching this

src/app/services/client/question.ts
```diff
@@ -18,6 +18,7 @@ import {
 } from "firebase/firestore";
 import { IdentifiableQuestion, IdentifiableQuestions } from "@interfaces/type";
 
+/* Why not pass in a Question type */
```
wkim10, line 21: Oh yeah @nickbar01234 @lnguyen2693  Johnny and I were wondering what the benefit is in creating the AddQuestion type? Can't we just pass in a Question type directly since we're just adding a new Question?

src/app/components/board/QuestionDetails.tsx
locmai1: I think it should be fine for now since we currently only have Student View and TA View and if this component is used in other places in the future, something we could do is create styling functions that take in what type of view it is and determine the sx styling which we can pass into the component without ternaries.

locmai1:
Looks fire, yall cooked 🙏

locmai1 (approved):

wkim10 (approved):
Mr. PM TL Tan is the goat, I made a few tiny changes to clean the code and also updated the "Create Announcement" button to follow the same style as the other fixed buttons. Besides that (and the small question I asked in question.ts) I think it looks great

src/app/services/client/question.ts
lnguyen2693: The initial idea was to make sure whatever passed in had at least those 5 fields of Question type (because I didn't want to care about the timestamp when I wrote that). But, we already control the `addQuestion()` call so yeah, you're right, it is not necessary anymore, we can just pass in a Question type directly

src/app/hooks/useCourseData.tsx
```diff
@@ -20,10 +27,6 @@ export const useCourseData = (props: CourseDataProps) => {
       students: [],
     };
   }
```
johnny-t06, line 28: @wkim10 I think we should have this check before the state initializations because it wouldn't be necessary for the !fetchUsers case . This case is used a lot more than the fetchUsers case, so I don't want to have unnecessary states.

wkim10: @johnny-t06 That's a good catch, I actually didn't really think about it when I was moving that if statement down. But I just looked it up and it seems like it's not good practice to return before any hooks (conditionally rendering hooks):

https://react.dev/reference/rules/rules-of-hooks
https://stackoverflow.com/questions/54938236/can-you-early-return-with-react-hooks

src/app/services/client/question.ts
wkim10: Ahh gotcha that makes sense. Sounds good!

src/app/hooks/useCourseData.tsx
johnny-t06: Ah, great catch!

Issue #15, opened 2024-08-05
 Queue Page


johnny-t06:
![image](https://github.com/user-attachments/assets/f72cf767-9b22-4a8a-bab5-6d357cd23751)
![image](https://github.com/user-attachments/assets/8483ecf4-a3d2-4771-b7ca-d7b20bfcc984)
## Description
Implement the Queue page with the following states + Leave Queue functionality
- Empty queue
- Closed off queue
- Normal queue
- Currently in queue
- Your turn!

Note:
Join Queue and Edit submission will be implemented as part of another ticket!

Issue #14, opened 2024-08-05
Join and Edit Form


johnny-t06:
## Description
As both forms are essentially the same, create a component to implement this form. 
Implement both front and backend interactions
![image](https://github.com/user-attachments/assets/524f911f-c65c-48a2-a707-6feb619912ab)

PR #38, opened 2025-01-09
Delete and Edit Submission

lnguyen2693:
# Description

The student now can:
- View their position on the queue
- Leave the queue (Delete their question)
- Edit their submission

P/s: There's no flag to make sure each student can only add 1 question yet, so for now if we add multiple questions, your position will be the first one you add 

Screenshot 2025-01-10 at 3 52 59 AM

## Test
1. Add yourself as a student of a class
2. Create a new question, edit question, leave queue
3. Create a new question again (after delete the first one)
4. Log into a different account as a student
5. Add a new question
6. Return to the initial account, edit question, leave queue


src/app/components/queue/EditQuestion.tsx
```diff
@@ -0,0 +1,138 @@
...
+            display: "flex",
+            flexDirection: "column",
+            rowGap: "12px",
+            justifyContent: "space-around",
+            alignItems: "center",
+          }}
+        >
+          {/* if you're at position 5 in the queue, does that mean there are 
+              4 queue ahead of you? Which number should I show here? */}
+          <Box sx={{ fontWeight: 700, fontSize: 57 }}>{queuePos}</Box>
```
lnguyen2693: In this case, I'm showing the number 5 now. It makes sense that I see my queue position there, but doesn't make sense when I read "queues ahead of you"

johnny-t06: Yea I agree its a little confusing, lets ask designers 👀

src/app/services/client/course.ts
```diff
@@ -177,4 +177,4 @@ export const deleteCourse = async (courseID: String) => {
 
   await deleteDoc(courseDoc);
   console.log("Course deleted");
```
johnny-t06: nit: Can remove comments

johnny-t06 (changes requested):
The flow and designs looks great. There's a bug when you edit your submission, save, and then try to edit again -- the form is not populated.

lnguyen2693:
That's weird because I tested it. How do I reproduce the bug?

johnny-t06:
I add a question and then edit it once. Then I try to edit again but the fields aren't populated. 
Also can we render the button as "edit submission" or what's on the figma instead of the "Join now" at bottom of the form?



https://github.com/user-attachments/assets/a1ffa16b-6e5f-4e6b-9f5d-0a8f8b441ed0

johnny-t06 (approved):
LGTM! Looks great, feel free to merge! I added a few styling changes and added a mui modal prop for an aria-hidden error. Linking to #14 and #15.

PR #39, opened 2025-01-10
UserContext Redirect + Middleware Auth validation

johnny-t06:
# Description
We were getting user from useUserSession hook or getUserSessionOrRedirect. However, we were getting React hook issues when using the router.push without the useEffect. 
1. Here I satisfy typescript by checking if user is null on pages bc the getUserSessionOrRedirect hook returns `IdentifiableUser | null` 
2. It also helps to render a blank screen as static paths like `/private/course/join` are cached / pre-fetched by Next.js so the following scenario happens. 
When navigating to a static page while signed out -- you could see the page for a split second before being redirected to `/`.

Middleware added to verify if user has a valid session cookie. Refer to [here](https://firebase.google.com/docs/auth/admin/manage-cookies#verify_session_cookie_and_check_permissions) for firebase's functions to use.

Also added some user information and log out button in the sidebar. It's not part of designs but I wanted easier access to both when testing session routing and I think it looks neat!

## Screenshots
image

## Test
Log in and click around different pages to make sure routing and loading looks.
Log out and try to visit any route, making sure it re-routes to `/login`
Test the buttons in the Sidebar.


firebaseAdmin.ts
```diff
@@ -0,0 +1,11 @@
+import * as admin from "firebase-admin";
+
+if (!admin.apps.length) {
+  admin.initializeApp({
+    credential: admin.credential.cert(
+      JSON.parse(process.env.FIREBASE_ADMIN_SDK_JSON || "{}")
```
wkim10: Similar to something Nick said when I initially added SendGrid, but maybe we can add a check here instead of falling back to {} (e.g. throwing an error if the env variable isn't defined). I'm not sure how initializeApp works under the hood but adding the check might cause the program to fail faster without attempting any Firebase operations that are done in initializeApp, and the error message might be more specific/user-friendly than whatever Firebase throws

src/app/login/page.tsx
```diff
@@ -1,11 +1,26 @@
...
+  const { user, onSignIn } = useUserSession();
+  const router = useRouter();
+  React.useEffect(() => {
+    if (user) {
+      router.replace("/private/course");
+    }
+  }, [user]);
+  if (user) {
+    return (
+      <Box className="flex h-screen w-screen flex-col items-center justify-center">
```
wkim10, line 19: I wonder if we should have another Spinner component that has this styling already added to it (the flex h-screen h-width...) since we're using it a lot and I'm assuming we'll keep using it more as we increase features

wkim10 (commented):
Looks really great, couldn't spot any issues with your logic and protecting private routes is working well. Thanks also for changing how we get the isUserTA bool. Just left two small comments!

firebaseAdmin.ts
johnny-t06: 100% agreed, will change!

PR #20, opened 2024-08-17
Won/home screen flow

wkim10:
#11 

Added pages + backend for "Join Class" as shown in the Figma section 1.

Testing:
- Go to /login and sign in, which takes you to /course
- Click "Join class" which takes you to /course/join
- Try inputting different types of class codes into the text field (test out edge cases)
- Can add new classes into Firebase or test using existing classes (with new "code" field)

Some images:
Screen Shot 2024-08-16 at 10 48 41 PM
Screen Shot 2024-08-16 at 10 46 26 PM
Screen Shot 2024-08-16 at 10 46 52 PM
Screen Shot 2024-08-16 at 10 47 09 PM

PR #40, opened 2025-01-11
Unify schema for user

nickbar01234:
# Description

Prior to this change, account created before #20 will not have `usernames` field in the document. As such, on first sign-in, they will see this error.

image

### DIFF
--- a/src/app/services/firestore.ts
+++ b/src/app/services/firestore.ts
@@ -24,7 +24,12 @@ export const userConverter = {
     return user;
   },
   fromFirestore(snapshot: QueryDocumentSnapshot, options: any): User {
-    return snapshot.data(options) as User;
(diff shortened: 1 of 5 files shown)


src/app/services/firestore.ts
```diff
@@ -24,7 +24,12 @@ export const userConverter = {
     return user;
   },
   fromFirestore(snapshot: QueryDocumentSnapshot, options: any): User {
-    return snapshot.data(options) as User;
+    const data = snapshot.data(options) as User;
+    return {
+      ...data,
+      role: data.role ?? "user",
```
nickbar01234, line 30: BTW, we should probably use an enum.

johnny-t06 (approved):
LGTM!

PR #41, opened 2025-01-11
Middleware + Verify Session Cookies

johnny-t06:
# Description
Prior to this, the middleware wasn't running because it was at the wrong file level. Firebase admin app did not follow a singleton pattern as recommended.
Also solved this with creation of verify session api + admin singleton.
image

Added verify session and log out session apis to handle cookies. 

There is a weird bug when signing in sometimes, you do not get redirected to `/private/course`

# Test
Log in and visit any routes, making sure you're not redirected to /login
Log out and make sure you can't visit any private routes.


firebaseAdmin.ts
```diff
@@ -1,14 +1,40 @@
-import * as admin from "firebase-admin";
-if (!process.env.FIREBASE_ADMIN_SDK_JSON) {
-  throw new Error("FIREBASE_ADMIN_SDK_JSON is not set");
-}
+import "server-only";
```
nickbar01234, line 1: Interesting library

package-lock.json
```diff
@@ -45,143 +45,73 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.24.7",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.24.7.tgz",
-      "integrity": "sha512-BcYH1CVJBO9tvyIZ2jVeXgSIMvGZ2FDRvDdOIVQyuklNKSsx+eppDEBq/g47Ayw+RqNFE+URvOShmf+f/qwAlA==",
+      "version": "7.26.2",
```
nickbar01234, line 48: How did you `package-lock` change but not your `package.json`? Am I missing something?

src/app/api/logout/route.client.ts
```diff
@@ -0,0 +1,11 @@
+export const logOutCookie = async () => {
```
nickbar01234: Why not simply `logout`? The implementation of using cookie should not be a factor.

package-lock.json
johnny-t06: Not sure, I might have uninstalled packages and deleted node_modules while testing

firebaseAdmin.ts
```diff
@@ -1,14 +1,40 @@
...
+            "\n"
+          ),
+          clientEmail: process.env.FIREBASE_CLIENT_EMAIL,
+        }),
+        projectId: process.env.NEXT_PUBLIC_FIREBASE_PROJECT_ID,
+        storageBucket: process.env.NEXT_PUBLIC_FIREBASE_STORAGE_BUCKET,
+      });
+    }
+  } catch (error) {
+    console.error("Error initializing Firebase Admin:", error);
```
nickbar01234: Should just throw here since it's unrecoverable?

src/middleware.ts
```diff
@@ -8,10 +7,20 @@ export const middleware = async (request: NextRequest) => {
     return NextResponse.redirect(new URL("/login", request.url));
   }
 
+  const origin = request.nextUrl.origin;
   try {
-    // Verify the session cookie with Firebase Admin SDK
-    await admin.auth().verifySessionCookie(sessionCookie.value, true);
+    const res = await fetch(`${origin}/api/session`, {
```
nickbar01234, line 26: Does `/api/session` by itself not work?

nickbar01234:
Given that we faced an infinite login earlier on, should we add any backdoor to get out of it in production? Some ideas

1. Mode to bypass requiring cookies to exist
2. If user is defined, but the middleware is rejected, should automatically logout to force the login flow again

nickbar01234:
Otherwise, the changes look sensible

package-lock.json
nickbar01234: Can you run `npm i` again locally? If you deleted packages, there should be a diff in `package.json`.

src/middleware.ts
johnny-t06: [This](https://nextjs.org/docs/messages/middleware-relative-urls) mentions it a little. Since middleware runs on server, I think it doesn't have the relative url access.

src/middleware.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const redirectToLogin = (request: NextRequest, hasCookie: Boolean) => {
+  const loginUrl = new URL("/login", request.url);
+  loginUrl.searchParams.set("loggedOut", "true");
+  const response = NextResponse.redirect(loginUrl);
+  if (hasCookie) {
+    response.cookies.delete("session");
+  }
+  return response;
+};
```
johnny-t06, line 10: On bad session token (i.e. DNE or not a valid token), we can reroute to /login?loggedOut=true. Then on /login and we check for loggedOut == true, to reset user context/trigger full logout.

johnny-t06:
In my last two commits, I added a env to bypass the session check in our middleware. I also added a re-route to /login in the middleware that would fully reset the User Context. I added a GetSession api, but didn't use it. Let me know if I should remove it!

PR #42, opened 2025-01-11
Persist session cookies

nickbar01234:
# Description

To be honest, I'm not quite sure what was wrong with your original approach - seems sensible if we're using Vanilla API. Also took some time to address a couple of bugs

- We should clear the cookies on signout instead of nullifying the expired time - this guarantees that the next time user log in, they won't have a valid session
- When we're trying to authenticate a session and the cookie is invalid, this is not possible to recover from right? So we should just delete it.

FYI, the [official documentation](https://firebase.google.com/docs/auth/admin/manage-cookies#sign_in) recommend that we also protect against CRSF attacks. I didn't address it in this PR since our App is not too critical.

### DIFF
--- a/src/app/api/auth/route.ts
+++ b/src/app/api/auth/route.ts
@@ -15,19 +15,14 @@ export const POST = async (request: NextRequest) => {
     const options = {
       maxAge: expiresIn,
       httpOnly: true,
-      secure: true, // Use secure cookies in production
+      secure: process.env.NODE_ENV === "production", // Use secure cookies in production
     };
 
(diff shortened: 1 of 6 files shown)


nickbar01234:
Seems like behavior is not consistent on Chrome 👀 . Will take another look - also did find a relevant issue here https://stackoverflow.com/questions/74303214/nextjs-middleware-cookies-issue

johnny-t06 (approved):
LGTM! Doesn't seem to be non-deterministic anymore

PR #45, opened 2025-01-20
Queue Notifications

wkim10:
# Description

Added email notifications using SendGrid when a student is ready to be helped by TAs and for students at the top of the queue. The notifications are triggered whenever a TA clicks "Start helping" or "Mark as missing" in QuestionDetails.tsx.

The "ready to be helped" notification is triggered always when TA clicks "Start helping."

The other case (for a user being number 1 on the queue) was a little trickier. There is a chance it is triggered when a TA clicks either "Start helping" or "Mark as missing," as that is the only time when a student will possibly move up in the queue. If the current question that was "started being helped" or "marked as missing" is at the top of the queue, then a notification will be sent to the second person on the queue, as the second student will move up to become number 1 on the queue. In any other case (meaning current question is not at the top of the queue), then the question that is at the top of the queue will not change so no notification needs to be sent. Let me know if this logic is confusing or if there are any edge cases that I didn't catch with this.

Also if you need SendGrid key for .env let me know :)

## Issues

:)

## Screenshots

Screenshot 2025-01-20 at 5 00 54 PM
Screenshot 2025-01-20 at 5 01 14 PM

## Test

Need to do thorough testing for all possible edge cases. Try marking a student as missing and marking them as being helped. Try this with groups of students (meaning notification should be sent to each student in the group). Try marking a student in the middle of the queue as missing/being helped (which wouldn't trigger the notification for a student at the top of the queue because there was no change).

## Possible Downsides

Maybe there is a better way using cloud functions to ensure no edge cases are missing (we could constantly listen for the top of the queue and whenever there is a change, we just send an email to the new person at position 1). Also, I left some comments in GitHub in the code where there could be improvements in code structure, feel free to critique :)

## Additional Documentations

:)

### DIFF
--- a/src/app/components/board/QuestionDetails.tsx
+++ b/src/app/components/board/QuestionDetails.tsx
@@ -88,6 +88,7 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
   const onMissingRemove = async () => {
     if (question.state === QuestionState.PENDING) {
       await sendTopUserNotif();
+      await sendEmail(getEmailTemplate("STUDENT_MISSING", users[0].email));
       await partialUpdateQuestion(question.id, courseId, {
         state: QuestionState.MISSING,
       });
--- a/src/app/components/board/QuestionDetails.tsx
+++ b/src/app/components/board/QuestionDetails.tsx
@@ -2,13 +2,9 @@
 
 import { IdentifiableQuestion, IdentifiableUsers } from "@interfaces/type";
 import { Avatar, Box, Typography } from "@mui/material";
-// import { formatTimeDifference, hasPassed, trimUserName } from "@utils/index";
 import {
   formatTimeDifference,
-  // getActiveQuestionsByState,
-  // getUserSessionOrRedirect,
   getEmailTemplate,
-  // getUserSessionOrRedirect,
   hasPassed,
   trimUserName,
 } from "@utils/index";
@@ -26,7 +22,6 @@ import Spinner from "@components/Spinner";
 import { CustomButton } from "@components/buttons/CustomButton";
 import { QuestionState } from "@interfaces/db";
 import { useRouter } from "next/navigation";
-// import { serverTimestamp } from "firebase/firestore";
 import useApiThrottle from "@hooks/useApiThrottle";
 import { useUserOrRedirect } from "@hooks/useUserOrRedirect";
 import { serverTimestamp, Timestamp } from "firebase/firestore";
@@ -50,11 +45,6 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
   } = props;
 
   const user = useUserOrRedirect();
-  // const user = getUserSessionOrRedirect();
-  // const { questions } = useOfficeHour();
-  // const sortedQuestions = sortQuestionsChronologically(questions);
-  // const { [QuestionState.PENDING]: pendingQuestions } =
-  //   getActiveQuestionsByState(sortedQuestions);
   const router = useRouter();
   const [joinGroup, setJoinGroup] = React.useState<boolean>(
     question.group.includes(user!.id)
@@ -82,9 +72,7 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
     }
     setJoinGroup(!joinGroup);
   };
-  const { fn: throttledOnJoinGroup } = useApiThrottle({
-    fn: onJoinGroup,
-  });
+
   const onMissingRemove = async () => {
     if (question.state === QuestionState.PENDING) {
       await sendTopUserNotif();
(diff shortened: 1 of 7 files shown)


src/app/components/board/QuestionDetails.tsx
```diff
@@ -39,6 +43,10 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
   } = props;
 
   const user = getUserSessionOrRedirect();
+  const { questions } = useOfficeHour();
```
wkim10: Better way to do this? I didn't wanna get all the questions and do the filtering for pendingQuestions unless I really had to, but couldn't think of another way to handle the different cases (I'm only getting questions and pendingQuestions for use in sendTopUserNotif function)

src/app/components/board/QuestionDetails.tsx
```diff
@@ -80,6 +89,21 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
     }
   };
 
+  const sendTopUserNotif = async () => {
+    if (pendingQuestions.length > 1 && pendingQuestions[0].id === question.id) {
```
wkim10, line 134: Let me know if this if statement logic is confusing. Idea is that you only wanna send a notification that a user is at the top of the queue if you are altering the question state for a question that is currently at the top of the queue (as that would change the question at the top of the queue to the question that is second on the queue). In any other case, the question at the top would not change, so no need to send this notification

src/app/components/board/QuestionDetails.tsx
```diff
@@ -80,6 +89,21 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
...
+      pendingQuestions[1].group.forEach(async (userId) => {
+        const user = await getUser(userId);
+        if (user) {
+          await sendEmail({
```
locmai1: maybe we could have an emailTemplate that takes in a user email, this way we can also send custom notifs for missingPerson as well

src/app/components/board/QuestionDetails.tsx
locmai1: Mmmm, I think we could make a new function in `client/questions.ts` where we make a batch call and we could pass in `courseId` and `batchSize` and that function could also sort by timestamps as well. Something along the lines of this:

```js
export const getBatchedQuestions = async (courseId: string, batchSize: number, stateFilter: string) => {
  const questionsQuery = query(
    collection(db, `courses/${courseId}/questions`).withConverter(questionConverter),
    where("state", "==", stateFilter),
    orderBy("timestamp"),
    limit(batchSize)
  );

  const snapshot = await getDocs(questionsQuery);
  const questionsDocs: IdentifiableQuestions = snapshot.docs.map((doc) => ({
    id: doc.id,
    ...doc.data(),
  }));

  return questionsDocs;
};
```

src/app/components/board/QuestionDetails.tsx
locmai1: Makes sense to me 😼.

Adding onto it, is there any point that you'd think we have an overload of emails to send, which it might be beneficial to handle email sending through parallelization

src/app/components/board/QuestionDetails.tsx
```diff
@@ -236,11 +260,19 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
...
                   await partialUpdateQuestion(question.id, courseId, {
                     state: QuestionState.IN_PROGRESS,
                     helpedBy: user.id,
                     helpedAt: serverTimestamp(),
                   });
+                  users.forEach(async (user) => {
+                    await sendEmail({
+                      email: user.email,
+                      subject: "The TAs are ready to help!",
+                      body: "The TAs are ready to help you now! Please listen for your name.",
```
lnguyen2693: Should we have a separate notification message for other people in the group? Like "The TAs are ready to help a question group that you join" or something.

lnguyen2693 (commented):
Goat!! Love receiving emails from OfficeMins🤩
The deployment still failed though. Also the `QuestionDetails.tsx` becomes quite messy with the sending notification. Considering that we still have some other notification cases, I think it's better to separate notification stuff into a different file.

locmai1:
> Goat!! Love receiving emails from OfficeMins🤩 The deployment still failed though. Also the `QuestionDetails.tsx` becomes quite messy with the sending notification. Considering that we still have some other notification cases, I think it's better to separate notification stuff into a different file.

@wkim10 @johnny-t06 could deployment failing be due to sendGrid API key not in prod `env`?

wkim10:
> > Goat!! Love receiving emails from OfficeMins🤩 The deployment still failed though. Also the `QuestionDetails.tsx` becomes quite messy with the sending notification. Considering that we still have some other notification cases, I think it's better to separate notification stuff into a different file.
> 
> @wkim10 @johnny-t06 could deployment failing be due to sendGrid API key not in prod `env`?

Yeah Johnny and I checked and that's why deployment's failing

src/app/components/board/QuestionDetails.tsx
wkim10: Ohhh interesting yeah that might be less confusing

src/app/components/board/QuestionDetails.tsx
```diff
@@ -236,11 +266,27 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
...
                   });
+                  users.forEach(async (user, index) => {
+                    if (index === 0) {
+                      await sendEmail({
+                        email: user.email,
+                        subject: "The TAs are ready to help!",
+                        body: "The TAs are ready to help you now! Please listen for your name.",
+                      });
+                    } else {
+                      await sendEmail({
```
wkim10: This is separate email for other group members that Linh suggested, gonna need help testing since I need a group :)

src/app/services/client/question.ts
```diff
@@ -141,3 +146,29 @@ export const deleteQuestion = async (courseId: string, questionId: string) => {
...
+
+  const snapshot = await getDocs(questionsQuery);
+
+  const questionsDocs: IdentifiableQuestions = snapshot.docs.map((doc) => ({
+    id: doc.id,
+    ...doc.data(),
+  }));
+
+  return questionsDocs;
+};
```
johnny-t06, line 190: Very cool filtering Won!

johnny-t06 (approved):
LGTM Won!

src/app/services/client/question.ts
wkim10: Thank you Johnny!

src/app/components/board/QuestionDetails.tsx
```diff
@@ -66,9 +70,13 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
     }
     setJoinGroup(!joinGroup);
   };
-  
+
   const onMissingRemove = async () => {
     if (question.state === QuestionState.PENDING) {
+      await sendTopUserNotif();
+      await sendEmail(
+        getEmailTemplate("STUDENT_MISSING", users[0]?.email ?? "")
```
johnny-t06, line 80: Added a STUDENT_MISSING email with a safe guard just in case since it depends on existence of the group.

johnny-t06 (approved):
LGTM, great job won!

src/app/components/board/QuestionDetails.tsx
wkim10: Just tested this works great

PR #60, opened 2025-01-25
Leave queue functionality for private/group questions

### LINKED ISSUE #57: [Bug] Leave queue missing logic
# Description
'Leave queue' button does not contain logic for the question's author leaving a group. Simply plug in the logic of 'leave group' in board

Image
Image

locmai1:
# Description

Added functionality to leave queue button on queue page for both private/group question:

- Leaving queue with private question (1 user/author) should delete question
- Leaving queue with group question  (1 user/author will delete question and 2+ users will just remove user)

## Issues

#57 

## Screenshots

Leaving queue with private/group question with 1 user:

image

image

Leaving queue with group question with 2+ users:

image

Leaving queue with group question with 1 user:

image

image

## Test

Detailed in Screenshots 😼

## Possible Downsides

None that I can think of right now 😼. Let me know if I'm missing a case though!

## Additional Documentations

None

### DIFF
--- a/src/app/services/client/question.ts
+++ b/src/app/services/client/question.ts
@@ -106,6 +106,14 @@ export const leaveQuestionGroup = async (
   await updateDoc(questionDoc, {
     group: arrayRemove(userId),
   });
+
+  const updatedGroup = question.group.filter((id) => id !== userId);
+
+  if (updatedGroup.length === 0) {
+    await deleteQuestion(courseId, question.id);
+    return;
+  }
+
   return { ...question, group: question.group.filter((id) => id !== userId) };
 };
 
--- a/src/app/services/client/question.ts
+++ b/src/app/services/client/question.ts
@@ -17,6 +17,7 @@ import {
   arrayRemove,
 } from "firebase/firestore";
 import { IdentifiableQuestion, IdentifiableQuestions } from "@interfaces/type";
+import { runTransaction } from "firebase/firestore";
 
 /* Why not pass in a Question type */
 type AddQuestion = Pick<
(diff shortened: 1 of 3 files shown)


locmai1:
Moved deletion logic to DB since it is the same for both leave group and leave queue. 

Additional testing was done on `/board/[questionId]` concurrently between users to test router redirecting to `/board` upon leaving group.

lnguyen2693 (approved):
lgtm!

src/app/services/client/question.ts
```diff
@@ -106,6 +106,14 @@ export const leaveQuestionGroup = async (
...
     group: arrayRemove(userId),
   });
+
+  const updatedGroup = question.group.filter((id) => id !== userId);
+
+  if (updatedGroup.length === 0) {
+    await deleteQuestion(courseId, question.id);
+    return;
+  }
+
```
johnny-t06, line 124: Can we perform the group check before the update? Aka if the group was equal to 1, then we can delete the question - otherwise we can normall arrayRemove? We can save a read-write here.
(There is a possible race condition if someone joins at the exact moment, but its very minimal)

johnny-t06 (approved):
LGTM, small comment on api then feel free to merge!

locmai1:
Turned leaving queue/group into an atomic function.

Testing done by simulating delay in function and leaving on 2 users at relatively same time (within 2 secs).

Documentation: https://firebase.google.com/docs/firestore/manage-data/transactions

Downside: could this incur reads since transaction polls calls until one completes for another to run?

PR #61, opened 2025-01-25
Disable join queue for user already in the queue

lnguyen2693:
# Description
- Disable `join queue` button when the user is already in the queue
## Issues
#52  

## Screenshots
- User 1 (already in queue):

Screenshot 2025-01-25 at 12 12 03 AM

- User 2 (not in queue but join user 1's group):
Screenshot 2025-01-25 at 12 11 37 AM
Screenshot 2025-01-25 at 12 11 46 AM

## Test
- Use account_1 to join the queue and check `join queue` button
- Use account_2 to join account_1's question group , check account_2's `join queue` button to make sure it's not disabled

PR #47, opened 2025-01-22
Validation - Tufts Emails, useApiThrottle, Question Form 

johnny-t06:
# Description
In this PR, there are a few validation changes
- Restrict on Google Oauth provider for tufts.edu emails only.
- Validate the Question form for required fields. Red highlight on TextInputs / Error modal
- Added useApiThrottle to protect creating a question, joining a class, submitting feedback, joining a public group

## Screenshots
Not much front end changes!

## Test
Test logins with different accounts and different env settings! 
Test buttons to see if they're throttled
Test if disabling pop ups changes the login flow!

## Possible Downsides
N/A

PR #65, opened 2025-01-25
Fix Path Import

johnny-t06:
# Description
Current Vercel deployment is failing because #61 was merged in when I renamed the util getUserSessionOrRidirect to the useUserOrRedirect hook in #47.

## Test
Test if the Create Question works as intended!


locmai1 (approved):
LGTM!

Issue #51, opened 2025-01-24
[UI] Home Screen for Classes (WAITING ON DESIGNS)


wkim10:
# Description

If students are already enrolled in a class, this screen should not pop up when they return to the website. It should be either the front page of their enrolled class or like a “Canvas page” of all their enrolled classes.

<img width="364" alt="Image" src="https://github.com/user-attachments/assets/1da79842-34a8-4173-a7f7-868068ca29ca" />

# Testing

Make sure it looks good :)

PR #68, opened 2025-01-27
Sidebar, Course Join Page, Public Questions

### LINKED ISSUE #59: [Bug] Manage Classes Routes to Undefined Route
# Description
When you click on the Manage Classes on the sidebar, it brings you the a `/profile/edit` page. Since editing profile is not implemented, we can route to either `/profile` or wait for #51 (so that you can go from private classes to the classes home page). 

Image

johnny-t06:
# Description
In this PR,
- Implemented /private/course per new designs with currently joined classes
- Sidebar modifications (logout button, gap spacing) per designers
- Changed Board -> Public and renamed to Public Questions
- Reroute Manage Classes to `/private` in the mean time until we set up edit classes
- New Commit: Added Editing Location & Hours and Manage Classes now push `/private/course`

## Issues
Solve #51 #59

## Screenshots
Before:
= Public Board

After:
CS160 Algorithms
= Public Questions

## Test
1. Test Empty Classes on `private/course`
2. Test User with Classses on `private/course`
3. Test if Sidebar navigation Manage Classes goes to correct route on empty class / and non-empty class
4. Test Sidebar functionality

### DIFF
--- a/src/app/components/DisplayLocHours.tsx
+++ b/src/app/components/DisplayLocHours.tsx
@@ -1,0 +1,105 @@
+"use client";
+import { useOfficeHour } from "@hooks/oh/useOfficeHour";
+import { Box, Button, TextField, Typography } from "@mui/material";
+import { partialUpdateCourse } from "@services/client/course";
+import React from "react";
+import theme from "theme";
+
+interface DisplayLocHoursProps {
+  location: string;
+  editable?: boolean;
+}
+
+const renderWithLinks = (text: string) => {
+  const urlRegex = /(https?:\/\/[^\s]+|www\.[^\s]+)/g;
+
+  const parts = text.split(urlRegex);
+  return parts.map((part, index) => {
+    if (urlRegex.test(part)) {
+      const href = part.startsWith("http") ? part : `https://${part}`;
+      return (
+        <a
+          key={index}
+          href={href}
+          target="_blank"
+          rel="noopener noreferrer"
+          style={{ color: "#38608F", textDecoration: "underline" }}
+        >
+          {part}
+        </a>
+      );
+    }
+    return part;
+  });
+};
+
+export const DisplayLocHours = (props: DisplayLocHoursProps) => {
+  const { location, editable = false } = props;
+  const [isEdit, setIsEdit] = React.useState<boolean>(false);
+  const [currLocation, setCurrLocation] = React.useState<string>(location);
+  const [error, setError] = React.useState<string>("");
+  const { course } = useOfficeHour();
+
+  React.useEffect(() => {
+    setCurrLocation(location);
+  }, [location]);
+
+  const onEditDone = async () => {
+    if (isEdit && currLocation.trim() === "") {
+      setError("Location & Hours cannot be empty");
+      return;
+    }
+
+    if (isEdit && currLocation.trim() !== location) {
+      await partialUpdateCourse(course.id, { location: currLocation.trim() });
+    }
+    setIsEdit(!isEdit);
+  };
+
+  return (
+    <Box>
+      <Box className="flex flex-row justify-between">
+        <Typography
+          variant="h6"
+          color={theme.palette.text.primary}
+          sx={{ fontWeight: "bold" }}
+        >
+          Location & Hours
+        </Typography>
+        {editable && (
+          <Button onClick={onEditDone} sx={{ textTransform: "none" }}>
+            <Typography color={"#38608F"} variant="subtitle2">
+              {isEdit ? "Save" : "Edit"}
+            </Typography>
+          </Button>
+        )}
+      </Box>
+      <Box
+        sx={{
+          padding: "16px 16px",
+          gap: "10px",
+          flexDirection: "column",
+          display: "flex",
+          marginTop: "16px",
+        }}
+      >
+        {isEdit ? (
+          <TextField
+            focused
+            multiline
+            label="Location & Hours"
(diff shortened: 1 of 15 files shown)


johnny-t06:
image

image

Added:
Editing Location and Hours + Manage Classes now push `/private/course`

johnny-t06:
New commit:
image
- Made Loc & Hours the same as Announcements -- Same component for the Rounded Tile
- Made the whole Question Tile in `/board` clickable

src/app/components/EnrolledCourses.tsx
```diff
@@ -0,0 +1,82 @@
...
+      }
+      setFetching(false);
+    };
+    fetchCourses();
+  }, [user]);
+
+  if (!user || fetching) {
+    return (
+      <Box className="flex flex-col h-screen items-center justify-center">
+        <Spinner width={64} height={64} />;
```
locmai1, line 35: I see something similar in Linh's PR so maybe we can make loading a separate component (can be for future thing though)

src/app/components/DisplayLocHours.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+const renderWithLinks = (text: string) => {
+  const urlRegex = /(https?:\/\/[^\s]+|www\.[^\s]+)/g;
+
+  const parts = text.split(urlRegex);
+  return parts.map((part, index) => {
+    if (urlRegex.test(part)) {
+      const href = part.startsWith("http") ? part : `https://${part}`;
+      return (
+        <a
```
locmai1, line 22: Could we possibly use MUI Link here? https://mui.com/material-ui/react-link/

src/app/components/Sidebar.tsx
```diff
@@ -98,7 +94,7 @@ const Sidebar = () => {
           <Typography variant="h6" color={theme.palette.text.primary}>
             My Classes
           </Typography>
-          <MenuButton isOpen={false} />
+          {!isOpen && <MenuButton isOpen={false} />}
```
locmai1: Is this necessary since the sidebar would be hidden? Or is this for the menu button on the sidebar to disappear on closing animation?

src/app/components/EnrolledCourses.tsx
wkim10: Will address this in my loading screens PR!

src/app/components/Sidebar.tsx
wkim10: Oh good point, the menu button on sidebar appears briefly when it does the closing animation but otherwise it's not there. Depending on design preference we should either completely remove it or keep it without the isOpen condition

src/app/components/DisplayLocHours.tsx
```diff
@@ -0,0 +1,102 @@
...
+import React from "react";
+import theme from "theme";
+import { RoundedTile } from "./RoundedTile";
+
+interface DisplayLocHoursProps {
+  location: string;
+  editable?: boolean;
+}
+
+const renderWithLinks = (text: string) => {
```
wkim10, line 14: Holy this regex is fire

src/app/components/DisplayLocHours.tsx
```diff
@@ -0,0 +1,102 @@
...
+        {editable && (
+          <Button onClick={onEditDone} sx={{ textTransform: "none" }}>
+            <Typography color={"#38608F"} variant="subtitle2">
+              {isEdit ? "Save" : "Edit"}
+            </Typography>
+          </Button>
+        )}
+      </Box>
+      <RoundedTile>
+        {isEdit ? (
```
wkim10, line 89: Maybe it'd be nice to have a Undo button like for Announcement? In case they accidentally change something

src/app/components/DisplayLocHours.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const [error, setError] = React.useState<string>("");
+  const { course } = useOfficeHour();
+
+  React.useEffect(() => {
+    setCurrLocation(location);
+  }, [location]);
+
+  const onEditDone = async () => {
+    if (isEdit && currLocation.trim() === "") {
+      setError("Location & Hours cannot be empty");
```
wkim10: Once the error shows up it doesn't seem to go away until you reload the page (like even if you refill it with text the error stays after you save)

src/app/components/Sidebar.tsx
johnny-t06: Designers wanted to remove the MenuButton when the sidebar is open.

src/app/components/DisplayLocHours.tsx
johnny-t06: Great catch!

src/app/components/Sidebar.tsx
wkim10: Which MenuButton? The one on the sidebar or the one on the home page? Kinda hard to explain, but this MenuButton is the one on the sidebar, so when you have the isOpen condition on it, it briefly displays it when you close the sidebar. If we just want to remove the MenuButton from the sidebar then we should get rid of this entirely

src/app/components/DisplayLocHours.tsx
wkim10: The button looks great, but it seems that when you press it it saves your changes instead of reverting back to the original text

wkim10 (approved):
Great work Mr. Tan, feel free to merge!

PR #76, opened 2025-01-28
Queue indicators for private & public questions

### LINKED ISSUE #54: [Bug] Queue Arrow Displaying (WAITING ON DESIGNS)
# Description

Clicking on the arrow on the queue should lead to the question description (right screen).

Make sure it’s only public questions that can navigate to question details and remove arrow for private questions.

Image

Also something to consider: Make the difference between private and public questions more clear. 
The 1st question is a public question; 2nd and 3rd are private questions — they look the same (looks like I can join all of them). 

Image

# Testing

Make sure that arrow is gone for private questions and it navigates successfully for public ones.

locmai1:
# Description

Added a clearer indicator for private and public questions per Figma designs. TA can see all questions, users who post a private question can see while others can't.

Additional:

- Changed styling of queue header and queue items to match Figma designs on all screens
- Keeping question title and users to "Private" and "Anonymous" and awaiting decision from designers for further updates
- Added hover pointer cursor for clearer distinction of queue item being clickable (only works for non-mobile tho)

## Issues

 #54 

## Screenshots

Before:

image

With changes:

image

## Test

Tested on different student screens + TA screen making sure that routing goes to question details

## Possible Downsides

None hopefully

## Additional Documentations

Figma styling:

image

### DIFF
--- a/src/app/components/queue/QueueItem.tsx
+++ b/src/app/components/queue/QueueItem.tsx
@@ -29,11 +29,6 @@ const QueueItem = (props: QueueItemProps) => {
   const isShowDetails =
     isUserTA || question.questionPublic || question.group[0] === user.id;
 
-  const privateTemplate = {
-    title: "Private",
-    user: "Anonymous",
-  };
-
   React.useEffect(() => {
     const fetchUsers = async () => {
       const fetchedUsers = await getUsers(question.group);
@@ -73,7 +68,7 @@ const QueueItem = (props: QueueItemProps) => {
           overflow="hidden"
           textOverflow="ellipsis"
         >
-          {isShowDetails ? question.title : privateTemplate.title}
+          {isShowDetails ? question.title : "Private"}
         </Typography>
         <Typography
           fontSize="14px"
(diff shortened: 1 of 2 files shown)


nickbar01234:
This feels a bit squished. Can we make the button one per line?

image

src/app/components/queue/QueueItem.tsx
```diff
@@ -68,11 +82,19 @@ const QueueItem = (props: QueueItemProps) => {
           overflow="hidden"
           textOverflow="ellipsis"
         >
-          {loading ? "Loading..." : users.map(trimUserName).join(", ")}
+          {loading
+            ? "Loading..."
+            : isShowDetails
+            ? users.map(trimUserName).join(", ")
+            : privateTemplate.user}
```
johnny-t06: We should still show the private user's name on the queue to other students, I don't think the anonymity would add to the experience.

johnny-t06:
> This feels a bit squished. Can we make the button one per line?
> 
> image

Seems like this doesn't happen for iPhone SE (one of the smallest phones), so we might be good.

johnny-t06 (approved):
LGTM, small comment

PR #78, opened 2025-01-28
Query by batch

nickbar01234:
# Description

A small improvement to how we fetch users. Note that I would want us to move to a context based implementation to avoid querying users at multiple level of the components. Alternatively, we want to add in a cache mechanism to avoid multiple network calls - something like react-query.

### DIFF
--- a/src/app/services/client/user.ts
+++ b/src/app/services/client/user.ts
@@ -1,9 +1,13 @@
 import {
+  collection,
   deleteDoc,
   doc,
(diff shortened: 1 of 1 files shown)


johnny-t06 (approved):
LGTM, replaced id with documentId() because we aren't storing id's in `/users` collection.

nickbar01234:
Ohh good catch, thanks!

PR #80, opened 2025-01-28
Merge state

nickbar01234:
# Description

Previously, `useLoadingValue` overrides the state completely each time. We may run into a case when value hasn't updated but snapshot is executed, effectively overriding the previous callback. Additionally, we used `snapshot.docs`, which includes all type of modifications. Luckily, we haven't ran into any problems yet, because questions are additive for now.

## Screenshots

Also took the chance to fix the UI for TA view history at `/private/course/cs160/board`

image


src/app/hooks/utils/useLoadingValue.tsx
```diff
@@ -6,18 +6,31 @@ export enum State {
...
-    setLoadingState({ state: State.SUCCESS, value: value });
+  const setLoadedValue = (value: T | ((data: T) => T)) => {
+    if (typeof value === "function") {
+      const fn = value as Function;
+      setLoadingState((prev) => {
+        const args = prev.state === State.SUCCESS ? prev.value : init;
+        return { state: State.SUCCESS, value: fn(args) };
+      });
+    } else {
+      return setLoadingState({ state: State.SUCCESS, value: value });
```
johnny-t06: nit: can remove the return to be consistent with the if case

src/app/hooks/oh/useQuestionsLoader.tsx
```diff
@@ -21,45 +23,51 @@ interface UseQuestionsLoaderProps {
...
+    init: [],
+  });
   const unsubscriber = React.useRef<Unsubscribe | null>(null);
 
   React.useEffect(() => {
-    getQuestions(props.courseId).then((value) => {
-      setValue(value);
-    });
-  }, [props.courseId]);
-
```
johnny-t06, line 33: Opted to remove this to favor the onSnapshot. It conflicted with the new docByType.

src/app/hooks/oh/useQuestionsLoader.tsx
```diff
@@ -21,45 +23,51 @@ interface UseQuestionsLoaderProps {
...
-    if (state.state !== State.SUCCESS) {
-      if (unsubscriber.current !== null) {
-        unsubscriber.current();
-        unsubscriber.current = null;
-      }
-      return;
+    if (unsubscriber.current !== null) {
+      unsubscriber.current();
+      unsubscriber.current = null;
     }
```
johnny-t06, line 35: I don't think we need to return on !State.Success because we can't fetch with onSnapshot (we'll be stuck in loading).

src/app/hooks/oh/useQuestionsLoader.tsx
```diff
@@ -21,45 +23,51 @@ interface UseQuestionsLoaderProps {
...
-            id: doc.id,
-            ...data,
-            timestamp: data.timestamp ?? Timestamp.now(),
-          });
-        });
-
-        setValue(questionsList);
-      }
+      where("state", "!=", 3),
+      where("timestamp", ">", oneDayAgo)
```
johnny-t06, line 44: Added a query for live questions only!

johnny-t06 (approved):
LGTM, removed some state checks + added query for live questions only on onSnapshot level.

Feel free to remove the debug comments after testing.

src/app/hooks/oh/useQuestionsLoader.tsx
nickbar01234: why does it conflict?

johnny-t06: First, it would setValue say 100 questions into Loader. 

Then the onSnapshot hits and sees the same 100 questions because its considered a `docChange["added"]` (as this is the first iteration). Therefore, the loader state then holds a duplicate per question, one with `{...question, type: "added"}` and one just `{...question}`.

The problem stems from the removed useEffect - not being a onSnapshot and when we didn't have the following filter in the snapShot useEffect
` const addedDocs = added
          .filter((doc) => !prev.some((prevDoc) => prevDoc.id === doc.id))`

src/app/hooks/oh/useQuestionsLoader.tsx
```diff
@@ -21,45 +23,51 @@ interface UseQuestionsLoaderProps {
...
     );
 
+    const unsubscribe = onSnapshot(q, (snapshot) => {
```
wkim10, line 46: It seems that if I create a question, then edit it, then attempt to leave the queue, it doesn't live update the deletion part (if you reload then my question disappears)

src/app/hooks/oh/useQuestionsLoader.tsx
nickbar01234: Hmm, how do I reproduce this behavior? I turned it back on and I don't see any duplicates.

image

My original intention with the `useEffect` is to fetch once on load, since my understanding is that `onSnapshot` only runs when you're listener is active and data is mutated.

src/app/hooks/oh/useQuestionsLoader.tsx
nickbar01234: Correct, which is why we have the `useEffect` originally

src/app/hooks/oh/useQuestionsLoader.tsx
johnny-t06: Resolved offline - onSnapshot does provide initial data.

PR #82, opened 2025-01-28
Show 'you're next' and add student missing screen

### LINKED ISSUE #55: [UI] You're Next For Everyone
# Description

The “You’re Next” should appear to EVERYONE in the group, not just the person who made the question.

Image

# Testing

Try creating multiple groups where you're the leader vs you're just a member and see if it pops up for everyone.

### LINKED ISSUE #75: [UI] Student Display "You are Missing Screen"
# Description
When a student is marked as missing by the TA, display the following screen. Check the current question user is in and the QuestionState.
Image

### LINKED ISSUE #86: Change styling for position component
Some small styling changes

Image

lnguyen2693:
# Description
In this PR:
- Add student missing screen
- Make everyone in a group see the question's position in the queue
- Fix queue position off when there is missing queue or current helping queue

## Issues
Resolve #55, resolve #75, resolve #86

## Screenshots
- Student can see the position in queue of the question they join
Screenshot 2025-01-28 at 5 58 48 PM
Screenshot 2025-01-28 at 6 02 28 PM

- Student missing screen
Screenshot 2025-01-28 at 6 04 36 PM

## Test
- Join a question to see if the queue position component shows up
- Join a question and create a question to see which question's position is shown in the position queue component

### DIFF
--- a/src/app/components/queue/EditQuestion.tsx
+++ b/src/app/components/queue/EditQuestion.tsx
@@ -1,17 +1,14 @@
 import { Box, Button, Container } from "@mui/material";
 import NotificationAddOutlinedIcon from "@mui/icons-material/NotificationAddOutlined";
 import KeyboardReturnOutlinedIcon from "@mui/icons-material/KeyboardReturnOutlined";
-import NotificationsActiveOutlinedIcon from "@mui/icons-material/NotificationsActiveOutlined";
 import { useOfficeHour } from "@hooks/oh/useOfficeHour";
 import React from "react";
 import { CustomModal } from "@components/CustomModal";
 import { leaveQuestionGroup } from "@services/client/question";
 import QuestionForm from "./form/QuestionForm";
 import { useUserOrRedirect } from "@hooks/useUserOrRedirect";
-import TaCard from "@components/tas/TaCard";
 import { useCourseData } from "@hooks/useCourseData";
 import Spinner from "@components/Spinner";
-import { QuestionDetails } from "@components/board/QuestionDetails";
 import { IdentifiableQuestion } from "@interfaces/type";
 import { QuestionState } from "@interfaces/db";
 import { StudentHelping } from "./StudentHelping";
@@ -128,7 +125,7 @@ export const EditQuestion = (props: editQuestionProps) => {
         }}
       >
         <Box sx={{ fontWeight: 700, fontSize: position === 0 ? 35 : 55 }}>
-          {position === 0 ? "You're Next!" : position}
+          {position === 0 ? "You're Next!" : position + 1}
         </Box>
         {position !== 0 && (
           <Box sx={{ fontWeight: 500, fontSize: 10, color: "#545F70" }}>
--- a/src/app/components/queue/EditQuestion.tsx
+++ b/src/app/components/queue/EditQuestion.tsx
@@ -124,14 +124,13 @@ export const EditQuestion = (props: editQuestionProps) => {
           alignItems: "center",
         }}
       >
-        <Box sx={{ fontWeight: 700, fontSize: position === 0 ? 35 : 55 }}>
+        <Box sx={{ fontWeight: 400, fontSize: '12px', color: "#545F70" }}>
+            Your queue position
+          </Box>
+        <Box sx={{ fontWeight: 700, fontSize: '57px' }}>
           {position === 0 ? "You're Next!" : position + 1}
         </Box>
-        {position !== 0 && (
-          <Box sx={{ fontWeight: 500, fontSize: 10, color: "#545F70" }}>
-            Queues ahead of you
-          </Box>
-        )}
+        
         <Box
           sx={{
             fontWeight: 400,
--- a/src/app/components/queue/EditQuestion.tsx
+++ b/src/app/components/queue/EditQuestion.tsx
@@ -34,36 +34,6 @@ export const EditQuestion = (props: editQuestionProps) => {
 
   const [leaveQueueModal, setLeaveQueueModal] = React.useState<boolean>(false);
 
-  // not in queue, not join any question,
-  // and not have any IN_PROGRESS or MISSING question
-  if (queuePos === -1 && groupPos === -1 && currQuestion.title === "") {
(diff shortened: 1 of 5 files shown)


src/app/components/queue/EditQuestion.tsx
```diff
@@ -45,78 +54,23 @@ export const EditQuestion = (props: editQuestionProps) => {
     );
   }
 
-  if (currQuestion.helpedBy !== "") {
+  if (currQuestion.state === QuestionState.IN_PROGRESS) {
     const ta = tas.find((myTa) => myTa.id === currQuestion.helpedBy);
-
     return (
```
nickbar01234: Can we have HTLM elements grouped together? This makes the code a bit hard to follow

src/app/components/queue/StudentHelping.tsx
```diff
@@ -0,0 +1,84 @@
...
+import TaCard from "@components/tas/TaCard";
+import {
+  IdentifiableCourse,
+  IdentifiableQuestion,
+  IdentifiableUser,
+} from "@interfaces/type";
+import NotificationAddOutlinedIcon from "@mui/icons-material/NotificationAddOutlined";
+import { Box } from "@mui/material";
+
+interface studentHelpingProps {
```
nickbar01234: nit: Upper case first letter

src/app/components/queue/EditQuestion.tsx
```diff
@@ -1,27 +1,28 @@
...
-import TaCard from "@components/tas/TaCard";
 import { useCourseData } from "@hooks/useCourseData";
 import Spinner from "@components/Spinner";
-import { QuestionDetails } from "@components/board/QuestionDetails";
 import { IdentifiableQuestion } from "@interfaces/type";
+import { QuestionState } from "@interfaces/db";
+import { StudentHelping } from "./StudentHelping";
+import { StudentMissing } from "./StudentMissing";
 
 interface editQuestionProps {
```
locmai1: nit: Capitalize here

src/app/components/queue/EditQuestion.tsx
```diff
@@ -145,16 +69,48 @@ export const EditQuestion = (props: editQuestionProps) => {
...
+  // and not have any IN_PROGRESS or MISSING question
+  if (queuePos === -1 && groupPos === -1 && currQuestion.title === "") {
+    return null;
+  }
+  const position =
+    queuePos === -1
+      ? groupPos
+      : groupPos === -1
+      ? queuePos
+      : Math.min(queuePos, groupPos);
```
locmai1, line 83: Not sure if this will happen but would there ever be a case where `queuePos === -1` and `groupPos === -1` but `currQuestion.title !== ""`? This case would fail the conditional right above and return -1

src/app/components/queue/EditQuestion.tsx
```diff
@@ -145,16 +69,48 @@ export const EditQuestion = (props: editQuestionProps) => {
...
+
+  if (loading) {
+    return (
+      <div className="flex items-center justify-center h-screen ">
```
locmai1: nit: Could we make the div a MUI Box? just for consistency

src/app/components/queue/EditQuestion.tsx
```diff
@@ -145,16 +69,48 @@ export const EditQuestion = (props: editQuestionProps) => {
...
+  if (currQuestion.state === QuestionState.MISSING) {
+    return <StudentMissing currQuestion={currQuestion} course={course} />;
+  }
+
   return (
     <>
-      <CustomModal
-        title="Leave queue?"
-        subtitle="You'll lose your place in line and
+      {position === queuePos && (
```
locmai1: nit: could we also consolidate this into a variable since its used a couple times?

src/app/components/queue/EditQuestion.tsx
```diff
@@ -216,20 +171,27 @@ export const EditQuestion = (props: editQuestionProps) => {
...
+              onClick={() =>
+                position === queuePos ? setLeaveQueueModal(true) : leaveGroup()
+              }
             >
               <KeyboardReturnOutlinedIcon style={{ fontSize: "14px" }} />
-              <Box>Leave queue</Box>
+              {position === queuePos ? (
+                <Box>Leave queue</Box>
+              ) : (
+                <Box>Leave group</Box>
```
locmai1: What's the difference between these two Box?

src/app/components/queue/EditQuestion.tsx
lnguyen2693: Yess. I think the logic here is a bit hard to follow, I'm not sure if there's any better way to do it, but basically:
`queuePos` : student's position in the pending queue, aka the position of their PENDING question
`groupPos` : position in the pending queue of the question group they join
`currQuestion`: the student's question, can be in PENDING, MISSING, or IN_PROGRESS state. The question group that they join is NOT `currQuestion`.
so the case where `queuePos === -1` and `groupPos === -1` but `currQuestion.title !== ""` is the case where a student doesn't join any question group and has one question that's currently in MISSING or IN_PROGRESS state. Both cases have their own student view screen so the `position === -1` shouldn't matter.

src/app/components/queue/EditQuestion.tsx
lnguyen2693: `Leave group` is for people who join the question, and `Leave queue` is for the author. There shouldn't be any difference in functionality. It just makes more sense for the user I think. Like if they haven't created any question, it would make more sense for them to leave a group instead of leave the queue since there's no queue for them to leave.

locmai1: Oh wait, I don't know why I thought these two were the same. I thought they were both "Leave queue" LOL. You can resolve this 😅

src/app/components/queue/EditQuestion.tsx
locmai1: Oh I see. Yeah, this makes sense then!

johnny-t06 (approved):
LGTM, looks great!!

PR #87, opened 2025-01-31
Filter by issue tags on board and history

### LINKED ISSUE #74: [Bug] Filter Public Questions By "Issue" Tags
Currently, it's filtering for the categories of the tags. We are making General tags static, and only Issue tags modifiable. So filtering by Issue tags makes the most sense

Image

locmai1:
# Description

Previously, we filtered by keys of tags that we used in the join queue form, now we want to filter by topics which is a part of the Issue key in course's tags field:

- Removed unnecessary sorting topics since I feel like it's not needed? (let me know tho, can revert)
- Added a new util function to get the topic tags for modularity

Other things addressed:

- Fixed an infinite loading + firebase invalid query bug on `board/history` when users load the history board
- Restructure `cs160` tags map in firebase to match `cs40` tags (Issue, General, How do you feel about the topic)

## Issues

#74 

## Screenshots

Board + History before:

image

image

Board + History after:

image

image

## Test

Tested current questions on Board + expiredQuestions on History for both CS160 and CS40

## Possible Downsides

None :)

## Additional Documentations

None :)


src/app/utils/index.ts
```diff
@@ -209,3 +209,7 @@ export const getEmailTemplate = (type: string, email: string) => {
       };
   }
 };
+
+export const getCourseTopicTags = (tags: Tags, tagKey: string): string[] => {
+  return (tags[tagKey].options || []).map((option) => option.choice);
+};
```
johnny-t06, line 234: Nice!

johnny-t06 (approved):
LGTM, can wait for discussions about the sorting

src/app/utils/index.ts
nickbar01234: nit: Prefer `??` over `||`, since `tags[tagKey].options` is undefined. See [thread](https://stackoverflow.com/a/61481631/19129136) for a discussion on when to use what

locmai1:
# Description

Previously, tags were ordered lexicographically. This PR also addresses the order of tags in both the join queue form as well as on the board, board/history, and displaying "How do you feel about the topic" only for TA view. Addressed:

- Tags are now ordered: General, Issue, How do you feel about the topic

Additional:

- Changed the text primary and secondary to reflect Figma
- Re-styled form to match Figma 
- Fixed the white border on tag selection in form

## Screenshots

Join queue form (before, after):

image

image


Board (user, TA):

image

image

Board history (user, TA):

image

image



## Test

All tests are done in photos above

## Possible Downsides

- Discovered a bug: TAs are join public board queues if they click on a question on the public board
- Right now, the TAG_ORDER is a constant but not sure if it needs to be dynamic or anything in future
- Joining queue form selection's label turn blue on filling form area but I think that's just MUI FormLabel

src/app/components/queue/form/QuestionForm.tsx
```diff
@@ -5,10 +5,13 @@ import {
   Drawer,
   FormControl,
   FormControlLabel,
+  FormLabel,
   IconButton,
   TextField,
+  Typography,
 } from "@mui/material";
 import CloseIcon from "@mui/icons-material/Close";
+// import FileUploadOutlinedIcon from "@mui/icons-material/FileUploadOutlined";
```
johnny-t06: nit: can remove

src/app/utils/index.ts
johnny-t06: ^^

johnny-t06 (approved):
LGTM

PR #88, opened 2025-02-03
fix: board + board history TA functionality, centralize error

locmai1:
# Description

Previously, TA can still join questions if they exist on the board. This PR introduces the `isUserTA` prop to `QuestionDetails`:

- adds extra layer of protection for all button functionality for QuestionDetails
- checks to display "Join group" and "Leave group"

Additionally:

- centralizes all global URL mismatch through a `not-found.tsx` file as per the documentation link below
- re-styled the board/[qid] and board/history/[qid] to match Figma styling

## Issues

Trust me 😭

## Screenshots

Before/after:

image

image

## Test

Tested user and TA view for question on board, history, board/qid, and history/qid. Users should have join/leave on board/qid and board and TA shouldn't. For history and history/qid, both users and TA do not have any buttons displayed.

## Possible Downsides

None that I know of so far...

## Additional Documentations

https://nextjs.org/docs/app/api-reference/file-conventions/not-found

### DIFF
--- a/src/app/components/board/QuestionDetails.tsx
+++ b/src/app/components/board/QuestionDetails.tsx
@@ -10,7 +10,7 @@ import {
 } from "@utils/index";
 import React from "react";
 import theme from "theme";
-import PeopleAltIcon from "@mui/icons-material/PeopleAlt";
+import PeopleAltOutlinedIcon from "@mui/icons-material/PeopleAltOutlined";
 import { getUser, getUsers } from "@services/client/user";
 import {
   getBatchedQuestions,
@@ -33,6 +33,7 @@ interface QuestionDetailsProps {
   fromTAQueue?: boolean;
   fromCurrentlyHelping?: boolean;
   fromStudentCurrentHelping?: boolean;
+  isUserTA: boolean;
 }
 
 export const QuestionDetails = (props: QuestionDetailsProps) => {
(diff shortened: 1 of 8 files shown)


src/app/components/board/QuestionDetails.tsx
```diff
@@ -319,10 +322,10 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
                 </CustomButton>
               )}
             </Box>
-          ) : fromStudentCurrentHelping ? null : (
+          ) : fromStudentCurrentHelping && isUserTA ? null : (
```
lnguyen2693, line 325: I don't think `&& isUserTA` is needed here since the `fromStudentCurrentHelping` is for the student view and having `&& isUserTA` means this will always return false

src/app/components/board/QuestionDetails.tsx
```diff
@@ -280,7 +283,7 @@ export const QuestionDetails = (props: QuestionDetailsProps) => {
                 Mark as done
               </CustomButton>
             </Box>
-          ) : fromTAQueue ? (
+          ) : fromTAQueue && isUserTA ? (
```
lnguyen2693, line 286: I'm not sure if `&& isUserTA` is needed here. I think `fromTAQueue` already means this is for TA view?

Issue #85, opened 2025-01-30
Loading Screens


wkim10:
# Description

Update loading screens to the following (goal is to have icon spin but only if you have time):

<img width="609" alt="Image" src="https://github.com/user-attachments/assets/20256e32-3ab0-4815-b84e-9a24e878a96b" />

# Testing

Make sure that loading screen looks good on all pages

PR #89, opened 2025-02-04
Loading Screens

wkim10:
# Description

Created new loading screen based on new logo spinning designs. Created new GlobalLoading and LoadingContext to centralize loading state.

## Issues

#85 

## Screenshots

PNG image

## Test

Try going through all the different flows and see if loading looks good

## Possible Downsides

Because all loading is now filling up the entire screen, I feel like it's sometimes a little abrupt. Let me know if this is the case for you, if it is I can create a new spinner component and add it individually to the different pages

## Additional Documentations

None :)


src/app/login/page.tsx
```diff
@@ -18,9 +17,35 @@ export default function Page() {
 
   if (user) {
     return (
-      <Box className="flex h-screen w-screen flex-col items-center justify-center">
-        <Spinner width={64} height={64} />
-      </Box>
+      <div className="fixed inset-0 flex items-center justify-center bg-white z-[100]">
```
wkim10, line 20: Wasn't able to figure out how to add loading state from context to login page. The issue is that the loading should stop when it redirects you after authenticating, but by using the loading state from the context, it doesn't ever stop loading unless you reload the page. This is a simple fix for now (just replacing the old spinner with the new one) but we can discuss possible fixes to this

src/app/hooks/useCourseData.tsx
```diff
@@ -13,9 +14,11 @@ export const useCourseData = (props: CourseDataProps) => {
...
   const { course } = useOfficeHour();
+  const { loading, setLoading } = useLoading();
+
   const [tas, setTAs] = React.useState<IdentifiableUsers>([]);
   const [students, setStudents] = React.useState<IdentifiableUsers>([]);
-  const [loading, setLoading] = React.useState(true);
+
   if (!user) {
     return {
       loading: false,
```
wkim10: Might not need to return "loading" as part of return if we're using centralized loading state with the context

src/app/components/queue/EditQuestion.tsx
```diff
@@ -23,26 +23,23 @@ interface editQuestionProps {
...
   const { queuePos, currQuestion } = props;
   const { course } = useOfficeHour();
-  const { tas, loading } = useCourseData({
-    fetchUsers: true,
-  });
-  const user = useUserOrRedirect();
-  if (!user) {
-    return null;
-  }
-
```
johnny-t06, line 34: I opted to not useCourseData because it fetches for students when we only need TAs here.

johnny-t06 (approved):
LGTM, looks great Won!

PR #92, opened 2025-02-04
Chunk Fetch for getUsers

johnny-t06:
# Description
Quick hotfix for getUsers fetching. The "IN" condition for a list of queries has a limit of 30, so this PR handles chunking the getUsers and also on the fetches 10 students on the `/course/[courseId]` page (rather than all users).

image

image

# Test
Go through the references of getUsers and see if they function the same.

### DIFF
--- a/src/app/services/client/user.ts
+++ b/src/app/services/client/user.ts
@@ -45,14 +45,29 @@ export const getUser = async (
 export const getUsers = async (
   userIds: string[]
 ): Promise<IdentifiableUsers> => {
-  const q = query(
-    collection(db, "users").withConverter(userConverter),
-    where(documentId(), "in", userIds)
-  );
-  const users = await getDocs(q);
-  return users.docs
-    .filter((doc) => doc.exists)
-    .map((doc) => ({ ...doc.data(), id: doc.id }));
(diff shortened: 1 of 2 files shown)


src/app/services/client/user.ts
```diff
@@ -45,14 +45,29 @@ export const getUser = async (
...
+  const chunks = [];
+  const chunkSize = 30;
+
+  for (let i = 0; i < userIds.length; i += chunkSize) {
+    chunks.push(userIds.slice(i, i + chunkSize));
+  }
+
+  const userPromises = chunks.map(async (chunk) => {
+    const q = query(
+      collection(db, "users").withConverter(userConverter),
```
locmai1, line 61: Instead of creating a new reference when processed in parallel, I think this can be a `const` outside of the promises

locmai1 (approved):
LGTM! Just a small comment but otherwise fire

src/app/services/client/user.ts
johnny-t06: It has to query per chunk, so I can't define on outside of the map. Maybe I'm missing something?

PR #98, opened 2025-02-11
bug: private question access restriction

### LINKED ISSUE #81: [Bug] Visit Course/Private Questions Through URL
# Description 
If userA makes a private post with id 123, the question details route would be `private/course/cs160/board/[123]` + `private/course/cs160/queue/[123]`. Any user can visit these routes and then join the group. 

Protect these routes with a front end check now and redirect, and Firebase security.

locmai1:
# Description

Previously, users are able to go to private question via URL regardless of if they made the question or not. This PR addresses that with:

- a hook that checks for question validity + user is a part of that question
- hook returns isUserTA, question, and loading state
- added board/history/[questionId] page

## Issues

#81 

## Screenshots

https://github.com/user-attachments/assets/f048c846-836f-414c-b3af-0f99ff138945

## Test

Tested the following on same course:

- 2 different users, one with private question, other trying to access private question through URL through queue
- single user, trying to access board question that exist & doesn't exist
- single user, trying to access board history question that exist & doesn't exist

## Possible Downsides

Not sure if this is the best way to handle it (please roast). Also not sure if I should be using loading context or not, lmk

## Additional Documentations

None!

### DIFF
--- a/src/app/hooks/oh/useQuestionAccessCheck.ts
+++ b/src/app/hooks/oh/useQuestionAccessCheck.ts
@@ -22,7 +22,7 @@ export const useQuestionAccessCheck = (
 
   useEffect(() => {
     if (!user || !course) {
-      throw new Error();
+      return;
     }
 
     if (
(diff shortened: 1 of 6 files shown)


johnny-t06 (approved):
LGTM! Will check with designers on the Spinner vs rendering null.

PR #103, opened 2025-02-25
Fixed reload on join class bug + text field height

### LINKED ISSUE #100: [Bug] Join Class Text Field
# Description

The "Join class" text field for the code is a little off on some devices, but it looks alright on others. Figure out why this is.

Image

# Testing

Make sure it looks consistent across devices :)

locmai1:
# Description

Previously, the height of the `/private/course/join` page's join class TextField would be cut in half upon render. This PR ensures that the TextField stays its fixed height on render. It also addresses a bug found while developing where if you refresh on the `private/course/join` page, you get an `application error` where there was a hooks render mismatch due to the `useUserSessionOrDirect`.

## Issues

#100 

## Screenshots

Before:

https://github.com/user-attachments/assets/7848deb0-ab30-48cf-8aea-95da5033d78b

After:

https://github.com/user-attachments/assets/7d0de42b-7192-4449-8c6b-c1466b7c49c2

## Test

Refer to videos above 😼

## Possible Downsides

Note that in second video of Screenshots, on reload, the TextField increases by like 1px. Not sure why that is though.

## Additional Documentations

None!

### DIFF
--- a/src/app/private/course/join/page.tsx
+++ b/src/app/private/course/join/page.tsx
@@ -20,10 +20,6 @@ const Page = () => {
   const [enrolledError, setEnrolledError] = React.useState(false);
   const [notFoundError, setNotFoundError] = React.useState(false);
 
-  if (!user) {
-    return null;
-  }
-
   const EnrolledButtons = [
     {
       text: "Ok",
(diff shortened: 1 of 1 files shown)


johnny-t06 (approved):
LGTM, that text box bug was nasty.

PR #104, opened 2025-02-28
Fix Join Group Ordering

### LINKED ISSUE #96: [Bug] Join Group Firebase ArrayUnion Not Ordered
# Description
Currently, we represent the group as an array in `Question` and use ArrayUnion to add userIds to the list. 
The problem is that Firebase's ArrayUnion does not guarantee order. It is more of like a Set structure, ensuring no duplicates. 

Note: I'm on the Vo L account
https://github.com/user-attachments/assets/9ee98a0e-deca-4be5-8867-026059796a5a

johnny-t06:
# Description

#92 introduced chunk fetching to get around Firebase's 30 query limit with the `where in` clause. 
However, chunk fetching with `where in` does not guarantee ordering of the results unless we sort it back ourselves (in respective to the input userIds).

Also small UI request from designers for the text below the tags
image

## Issues
Solves #96, it wasn't an arrayUnion issue as checking the Firebase when joining group appends to the end of the array.

## Test
- Try with 2+ accounts to see test joining and leaving conditions (ex: join end of group, promote to leader, leave and delete question)

### DIFF
--- a/src/app/components/board/index.tsx
+++ b/src/app/components/board/index.tsx
@@ -1,10 +1,17 @@
 import { IdentifiableQuestions } from "@interfaces/type";
-import { Box, Stack, ToggleButton, ToggleButtonGroup } from "@mui/material";
+import {
+  Box,
+  Stack,
+  ToggleButton,
+  ToggleButtonGroup,
+  Typography,
+} from "@mui/material";
 import Question from "./Question";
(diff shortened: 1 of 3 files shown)


wkim10 (approved):
You're so smart Johnny

locmai1 (approved):
LVGTM!

PR #106, opened 2025-02-28
Add Submission - schema + fixes

locmai1:
# Description

On #105, there is currently a bug where new lines on initial submission have `"(message added)"` next to them. This PR addresses that while also having better distinguishing between all added submissions using `Timestamp`.

Note: the Description field in the Question document is now an array of

```json
{
    "text": "string",
    "timestamp": "firebase Timestamp"
}  
```

## Issues

None

## Screenshots

https://github.com/user-attachments/assets/2d613d3d-2097-4a5a-a20a-bc283b61eb41

## Test

Created question and multiple added description to them as seen in video above

## Possible Downsides

None?

## Additional Documentations

None?

PR #105, opened 2025-02-28
Add Submission

wkim10:
# Description

Changed "Edit Submission" to "Add to Submission" feature

## Issues

#93 

## Screenshots

Screenshot 2025-02-28 at 2 18 43 AM
Screenshot 2025-02-28 at 2 18 55 AM
Screenshot 2025-02-28 at 2 19 09 AM

## Test

Try creating multiple questions and adding to them several times

## Possible Downsides

None :)

## Additional Documentations

None :)

PR #107, opened 2025-03-02
Edit tags

### LINKED ISSUE #72: Edit Tags
# Description

Allow TAs to be able to add/delete tags on the Home page

Image

# Testing

Try adding and removing many tags, make sure they show up properly when a student creates a question

locmai1:
# Description

Previously, there was no way to change the tags that are associated with a course. This PR adds the functionality to add & delete tags for a course and have it reflected in the Join Queue form as well as the board/history.

Note: added snackbar notifications (let me know if this should be removed)

## Issues

#72 

## Screenshots

https://github.com/user-attachments/assets/394f814c-9d09-4f06-a5fa-5becb3367965

## Test

Made a bunch of tags, verified it in the database, and checked that it was addable in both the join queue form as well as the board similar to video above.

## Possible Downsides

Might need to merge in #106 and #105 first since there is new structure of Description

## Additional Documentations

None!

### DIFF
--- a/src/app/components/DisplayIssueTags.tsx
+++ b/src/app/components/DisplayIssueTags.tsx
@@ -1,0 +1,239 @@
+import React, { useCallback, useState, useEffect } from "react";
+import { useOfficeHour } from "@hooks/oh/useOfficeHour";
+import { getCourseTopicTags } from "@utils/index";
+import {
+  Box,
+  Typography,
+  Button,
+  Chip,
+  TextField,
+  IconButton,
+  Snackbar,
+  Alert,
+  CircularProgress,
+} from "@mui/material";
+import theme from "theme";
+import CloseIcon from "@mui/icons-material/Close";
+import { addCourseTag, deleteCourseTag } from "@services/client/course";
+import useApiThrottle from "@hooks/useApiThrottle";
+
+interface DisplayIssueTagsProps {}
+
+const TAG_KEY = "Issue";
+
+export const DisplayIssueTags = (props: DisplayIssueTagsProps) => {
+  const { course } = useOfficeHour();
+  const topics = getCourseTopicTags(course.tags, TAG_KEY);
+  const [isEdit, setIsEdit] = useState<boolean>(false);
+  const [currentTag, setCurrentTag] = useState<string>("");
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [loadingTag, setLoadingTag] = useState<string>("");
+  const [error, setError] = useState<string>("");
+  const [showSnackbar, setShowSnackbar] = useState<boolean>(false);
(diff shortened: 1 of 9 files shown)


src/app/services/client/course.ts
```diff
@@ -214,3 +214,37 @@ export const deleteCourse = async (courseID: String) => {
...
+
+export const addCourseTag = async (courseID: String, tag: String) => {
+  const courseDoc = doc(db, `courses/${courseID}`).withConverter(
+    courseConverter
+  );
+
+  const newTag = {
+    choice: tag,
+    color: "orange",
+    colorFill: false,
```
lnguyen2693: The `color` and `colorFill` fields don't do anything for us anymore (we had them because of the initial design I believe). You can remove them here and in the interface!

lnguyen2693 (approved):
Only have one comment. Everything else looks good to me! 🐐

src/app/components/DisplayIssueTags.tsx
```diff
@@ -0,0 +1,243 @@
...
+  });
+
+  const handleKeyPress = useCallback(
+    (event: React.KeyboardEvent) => {
+      if (event.key === "Enter" && !fetching) {
+        throttledSubmitTag();
+      }
+    },
+    [throttledSubmitTag, fetching]
+  );
```
johnny-t06, line 89: Enter is a good idea but I think it wouldn't work on mobile, maybe we can submit button / logic - Wonkim025

src/app/components/DisplayIssueTags.tsx
```diff
@@ -0,0 +1,243 @@
...
+                </Typography>
+                {topic !== "Other" && (
+                  <IconButton
+                    onClick={() => handleDeleteTag(topic)}
+                    disabled={isLoading || loadingTag === topic}
+                  >
+                    {loadingTag === topic ? (
+                      <CircularProgress size={16} sx={{ color: "#43474E" }} />
+                    ) : (
+                      <CloseIcon fontSize="small" sx={{ color: "#43474E" }} />
```
johnny-t06, line 211: What is loadingTag in this case?

src/app/services/client/course.ts
```diff
@@ -214,3 +214,27 @@ export const deleteCourse = async (courseID: String) => {
...
 };
+
+export const addCourseTag = async (courseID: String, tag: String) => {
+  const courseDoc = doc(db, `courses/${courseID}`).withConverter(
+    courseConverter
+  );
+
+  await updateDoc(courseDoc, {
+    "tags.Tags.options": arrayUnion({
+      choice: tag,
```
johnny-t06, line 225: We should add note back

johnny-t06 (approved):
LGTM!

PR #108, opened 2025-03-05
New queue view

locmai1:
# Description

This PR implements new queue logic and views for students. Students view now has:
- Accordion dropdowns for questions on queue
- automatic queue placements based on the questions that they join/create
- handles priority in multiple group questions to show just the top

Note: add to submission is not implemented as it depends on #105 to be merged first

## Issues

https://www.figma.com/design/rGjxNWn0NenMKjGxNhC5i7/Office-Minutes-%5BSummer-2024%5D?node-id=2713-14372&t=V6HOHviOEDAPZFHy-0

## Screenshots

https://github.com/user-attachments/assets/40cb8bd2-002b-47a9-b80f-ff361df5037b

## Test

Create group and private questions. Join other group questions and test clicking dropdown and leaving and seeing the queue position update.

## Possible Downsides

As noted above: add to submission isn't working yet and pending other changes

## Additional Documentations

None

### DIFF
--- a/src/app/components/queue/EditQuestion.tsx
+++ b/src/app/components/queue/EditQuestion.tsx
@@ -70,7 +70,7 @@ export const EditQuestion = (props: EditQuestionProps) => {
 
   if (currQuestion.state === QuestionState.IN_PROGRESS) {
     const ta = tas.find((myTa) => myTa.id === currQuestion.helpedBy);
-    return ta ? <StudentHelping currQuestion={currQuestion} ta={ta} /> : null;
+    return ta && <StudentHelping currQuestion={currQuestion} ta={ta} />;
   }
 
   if (currQuestion.state === QuestionState.MISSING) {
--- a/src/app/components/queue/EditQuestion.tsx
+++ b/src/app/components/queue/EditQuestion.tsx
@@ -1,12 +1,9 @@
 "use client";
-import { Box, Button, Container, Typography } from "@mui/material";
-import NotificationAddOutlinedIcon from "@mui/icons-material/NotificationAddOutlined";
-import KeyboardReturnOutlinedIcon from "@mui/icons-material/KeyboardReturnOutlined";
+
 import { useOfficeHour } from "@hooks/oh/useOfficeHour";
 import React from "react";
 import { CustomModal } from "@components/CustomModal";
 import { leaveQuestionGroup } from "@services/client/question";
-import QuestionForm from "./form/QuestionForm";
 import { useUserOrRedirect } from "@hooks/useUserOrRedirect";
 import { IdentifiableQuestion, IdentifiableUsers } from "@interfaces/type";
 import { getUsers } from "@services/client/user";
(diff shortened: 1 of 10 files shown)


src/app/utils/index.ts
```diff
@@ -168,19 +168,19 @@ export const getQueuePosition = (
...
 
   const pendingPos = sortedPendingQuestions.findIndex(
-    (q) => q.group[0] === user.id
+    (q) => !q.questionPublic && q.group[0] === user.id
   );
   return {
     queuePos: pendingPos,
     groupPos: groupPos,
-    currQuestion: sortedActiveQuestions[position] ?? defaultQuestion(),
+    currQuestion: sortedActiveQuestions[pendingPos] ?? defaultQuestion(),
```
johnny-t06: pendingPos filters for sortedPendingQuestions and then accesses sortedActiveQuestions. This will never return a currQuestion that has a missing  state and therefore won't render the Missing Screen in the parent component.

johnny-t06: Marking missing for the group also doesn't display the screen, but I feel that's an expected case.

src/app/components/queue/EditQuestion.tsx
```diff
@@ -64,148 +62,61 @@ export const EditQuestion = (props: EditQuestionProps) => {
...
-
   if (currQuestion.state === QuestionState.IN_PROGRESS) {
     const ta = tas.find((myTa) => myTa.id === currQuestion.helpedBy);
-
-    if (!ta) {
-      return null;
-    }
-
-    return <StudentHelping currQuestion={currQuestion} ta={ta} />;
+    return ta ? <StudentHelping currQuestion={currQuestion} ta={ta} /> : null;
```
johnny-t06: nit: can use && here

src/app/components/queue/EditQuestion.tsx
```diff
@@ -64,148 +62,61 @@ export const EditQuestion = (props: EditQuestionProps) => {
...
-    >
-      Edit submission
-    </Button>
-  );
-
   // not in queue, not join any question,
   // and not have any IN_PROGRESS or MISSING question
   if (queuePos === -1 && groupPos === -1 && currQuestion.title === "") {
     return null;
   }
```
johnny-t06, line 83: This affects the missing screen rendering based on what is returned from the getQueuePos

johnny-t06 (commented):
Front end looks amazing, just small comments on how we should handle filtering of questions

src/app/utils/index.ts
johnny-t06: sortedActive has pending + missing 
problem is you have to maintain the actual queue positions because the missing shouldn't be factored in / affecting the queue order/numbering of the actual queue. When's someones marked missing, it's most likely (99%) at the top of the original queue.

johnny-t06 (approved):
LGTM Loc! Looks amazing, I added some extra logic for defaulting the Select Value for the drop down in QuestionForm.

